PR #38, opened 2016-08-02
Don't style the <a> anchor element

ErikNijland:
Third party plugins used anchors as well. In this case it was causing erros with Leaflet controls.


tietyk:
test this please

PR #242, opened 2016-11-01
Changed styling based on review nico

Kotrotsos:


modules/page/components/login-form/_login-form.scss
```diff
@@ -35,7 +39,7 @@
 .c-login-form__button {
   float: right;
   font-family: $header-font-family;
-  font-weight: 700;
+  font-weight: 500;
```
Slotkenov: Niet gewoon weghalen nu dan?

PR #284, opened 2016-11-23
Bug/tg 2135 column layout

boris-arkenaar:
Cleaned up dashboard columns
Also removed the extra pixel between menu item and menu, on Firefox.

### DIFF
--- a/modules/shared/shared.scss
+++ b/modules/shared/shared.scss
@@ -21,7 +21,6 @@
 @import "styles/utilities/aspect-ratio";
 @import "styles/utilities/clearfix";
 @import "styles/utilities/color";
-@import "styles/utilities/position";
 @import "styles/utilities/display";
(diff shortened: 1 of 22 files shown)


modules/shared/styles/utilities/_position.scss
```diff
@@ -0,0 +1,2 @@
+.u-position--relative { position: relative; }
+.u-position--absolute { position: absolute; }
```
Vegter: Kan weg

modules/shared/shared.scss
```diff
@@ -21,6 +21,7 @@
 @import "styles/utilities/aspect-ratio";
 @import "styles/utilities/clearfix";
 @import "styles/utilities/color";
+@import "styles/utilities/position";
```
Vegter: Kan weg

PR #292, opened 2016-11-28
Not using _display attribute anymore for search results of gebieden and

boris-arkenaar:
bouwblokken. Updated these tyes to display their subtype as well


boris-arkenaar:
Will recommit under a different branch name

PR #505, opened 2017-03-14
changed Jenkinsfile so that a branch doesn't get deployed on acc

KelvinPaska:


KelvinPaska:
@StephanZaat kan jij deze pull request reviewen?

PR #514, opened 2017-03-20
module-getter - enforce to reference modules with the getter syntax c…

Vegter:
…auses errors in the grunt files

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,6 +33,7 @@ module.exports = {
         "eol-last": "error",            //require or disallow newline at the end of files
         "no-shadow": "error",           // disallow variable declarations shadowing outer scope variables
         "angular/json-functions": "error",    // use angular.fromJson/toJson instead of JSON.parse and JSON.stringify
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
         "eol-last": "error",            //require or disallow newline at the end of files
         "no-shadow": "error",           // disallow variable declarations shadowing outer scope variables
         "angular/json-functions": "error",    // use angular.fromJson/toJson instead of JSON.parse and JSON.stringify
+        "angular/module-getter": "off", // When using a module, avoid variable, use chaining with the getter syntax
```
RRMoelker: Or consider only  disabling the rule on the following lines/files.

```
/atlas/grunt/babel.js
   9:9   error  Avoid using a variable and instead use chaining with the getter syntax  angular/module-getter
  97:13  error  Avoid using a variable and instead use chaining with the getter syntax  angular/module-getter

/atlas/grunt/tasks/build-tasks.js
  23:13  error  Avoid using a variable and instead use chaining with the getter syntax  angular/module-getter

```

PR #555, opened 2017-04-03
Feature/security fixes

boris-arkenaar:


boris-arkenaar:
Don't merge acc into new-design/master

PR #578, opened 2017-04-12
Feature/tg 3319 temp login

boris-arkenaar:


modules/shared/services/authenticator/authenticator.factory.js
```diff
@@ -34,7 +34,10 @@
```
RRMoelker: `applicationState` is not used

PR #636, opened 2017-05-23
Recover from error response, 404 is acceptable

boris-arkenaar:
The straatbeeld-thumbnail component handles error responses. A 404 is accepted.
A change was needed to the http-error-registrar to make local error handlers be executed before the global one and give the local error handlers the ability to recover (make the global error handler ignore it).


RobbinHabermehl:
test this please

PR #680, opened 2017-06-13
bower leaflet resolution no longer required [WACHT MET MERGEN, until after public release is stable]

RRMoelker:


RobbinHabermehl:
Sure? `bower   ECONFLICT Unable to find suitable version for leaflet` 😅

PR #749, opened 2017-07-25
Feature/tg 4175 catalogus api

boris-arkenaar:
* The `data-selection-api-ckan` factory uses a different `ROOT` than the other API endpoints
* The stripping and restoring of endpoint roots has been moved from the `state-url-converter` factory to the `state-url-conversion` constant which delegates it to the `uri-stripper` factory.
* Fixed a bug in the `detail-document-title` factory where it would crash if no glossary definition could be found.


jasperswart:
test this please

jasperswart:
test this please

PR #874, opened 2017-11-05
Update README.md

gokaygurcan:
- Remove unused requirement
- Add NODE_ENV variable

---

Hi,
It seems that Bower has been removed from the repository recently, but README has it as a requirement. Within this PR, it'll be removed as well.
Also, if you just use `npm start` as written in README file, you'll most likely end up with an error because of `babel-preset-react-app` package. It requires you to enter one of `development`, `test` or `production` values. Instead of adding it to the `start` script and forcing people to use `development` all the time, I think this is a better alternative.

I hope you'll accept this PR, thanks in advance. 

Best,
G.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,13 @@
 # City Data
 
 ## Requirements
-- NPM
-- Bower
+- npm
 - Webpack
 - Java JRE (needed to run protractor)
 
 ## Installations
 - npm install
-- npm start
+- NODE_ENV=development npm start
 - Open the server at http://localhost:8080/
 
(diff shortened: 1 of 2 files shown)


datapunt:

gokaygurcan:
Hi @jasperswart and @boris-arkenaar,
Actually I do not use it at all. We just moved to Amsterdam last week and I was checking what's on GitHub related to the city. Then I found the organizaton and checked the repositories. This one looked nice to contribute in the end and here we are now. 

Thanks again!

PR #946, opened 2017-12-18
DP-4993 losse eindjes url-state en visueel

boris-arkenaar:


modules/atlas/services/redux/reducer.factory.js
```diff
@@ -34,8 +34,8 @@ import { combineReducers } from 'redux';
                 user: UserReducer
             });
             const filteredState = {
-                ui: oldState.ui,
-                user: oldState.user
+                ui: deprecatedState.ui,
+                user: deprecatedState.user
```
boris-arkenaar: Dat was de fix?! :scream:

RRMoelker: afraid so

PR #963, opened 2018-01-08
Add a lastYear instead of the currentYear to fix that 2018 is shown

aronmartin:


datapunt:

PR #981, opened 2018-01-11
Add fix for scroll-wrapper - IE

aronmartin:


datapunt:

PR #985, opened 2018-01-15
DP-5095 pano18 spike

rmaijers:
todo : check if additional tests about 2018 are needed in hotspot.component.test.js


modules/shared/assets/images/hotspots/2018-hover.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
RRMoelker: Svgo?

PR #1014, opened 2018-01-31
Bugfix/dp 5211 map search results

RRMoelker:
Map search results reducer bug fixed and tested


src/map/ducks/search-results/map-search-results.js
```diff
@@ -14,10 +14,7 @@ export default function MapSearchResultsReducer(state = initialState, action) {
       return { ...state, isLoading: true, mapSearchResultsError: null };
 
     case FETCH_MAP_SEARCH_RESULTS_SUCCESS: {
-      const locationId = Object
-        .keys(action.location)
-        .map((key) => action.location[key])
-        .toString();
+      const locationId = Object.values(action.location).toString();
```
aronmartin: Beter! 👍

PR #1059, opened 2018-03-05
DP-5351 Fixed alignment of preview panel in print mode.

boris-arkenaar:
Also fixes DP-5146 (preview panel always displayed with full hight in
print mode).


boris-arkenaar:
test this please

PR #1075, opened 2018-03-12
Back to vh, as the % breaks on IE and FF

vjawala:


src/map/components/detail-result/_map-detail-result.scss
```diff
@@ -75,7 +75,7 @@ $below-list-height: 1px + $map-preview-panel-bottom;
     overflow-y: auto;
 
     .c-dashboard__body--error & {
-      max-height: calc(100% - #{$above-list-height + $below-list-height + $dashboard-error-height}); //NB: fix for IE
+      max-height: calc(100vh - #{$above-list-height + $below-list-height + $dashboard-error-height}); //NB: fix for IE
```
aronmartin: We just had a big refactor because 100vh is not equal 100% of the view height on devices. I think this change will revert that fix

vjawala: ah, clear. I updated it, it should be failsafe now

src/map/components/detail-result/_map-detail-result.scss
```diff
@@ -76,6 +76,12 @@ $below-list-height: 1px + $map-preview-panel-bottom;
 
     .c-dashboard__body--error & {
       max-height: calc(100% - #{$above-list-height + $below-list-height + $dashboard-error-height}); //NB: fix for IE
+
+      // use percentage on mobile, because some mobile devices don't handle vh correctly
+      // use vh on the rest, because in IE and FF the 100% gets calculated on the parent element and not on the window
+      @media screen and (min-width: $breakpoint-sm) {
```
aronmartin: This doens't seem like a proper solution. The 100vh issue is not only on < 539px screens. (Ipad, phone landscape etc)

Maybe you could debug why the 100% max-height is not working? Is max-height: 100% working when `.c-dashboard__body--error` is not appended?

PR #1105, opened 2018-04-04
DP-5401 Not sending auth headers when requesting CMS data

boris-arkenaar:


modules/shared/services/api/api.factory.test.js
```diff
@@ -136,6 +136,18 @@ describe('The api factory', function () {
         $httpBackend.flush();
     });
 
+    it('does not add an Authorization header if specified not to', function () {
+        // Not logged in
+        isLoggedIn = true;
```
vjawala: `// not logged in` -> `isLoggedIn = true;` ?

boris-arkenaar: Nice one :+1:

PR #1109, opened 2018-04-05
Feature/bac 881 aansluiten catalogus op frontend

aditudorache:
This feature is not yet completed but feedback would be welkome


aditudorache:
The detail page is now ready for a review so I dismiss this pull request and open a new one

boris-arkenaar:
@aditudorache Why not just keep using this PR?

aditudorache:
My mistake I think, I'll keep it open next time.

PR #1152, opened 2018-05-16
fix the unit test (only calling render, if the element to render into…

vjawala:
… is found). This is the way it is also done in other parts of the app

### DIFF
--- a/modules/header/components/site-header/site-header.component.test.js
+++ b/modules/header/components/site-header/site-header.component.test.js
@@ -3,7 +3,8 @@ describe('The dp-site-header component', () => {
         $rootScope,
         $window,
         origAuth,
-        mockedUser;
+        mockedUser,
+        $timeout;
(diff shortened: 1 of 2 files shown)


modules/header/components/site-header/site-header.component.test.js
```diff
@@ -55,6 +58,40 @@ describe('The dp-site-header component', () => {
...
+            };
+            spyOn($window.document, 'querySelector').and.returnValue(fakeCandidate);
+            spyOn($window, 'render').and.callThrough();
+            getComponent('', 'short');
+
+            $timeout.flush();
+            expect($window.render).toHaveBeenCalled();
+        });
+
+        it('does not do the react createElement call', () => {
```
boris-arkenaar: It does, or it doesn't. Which one is it? (Be more descriptive.)

PR #1173, opened 2018-05-30
Fix/wkpb legend

rmaijers:


rmaijers:
replaced by a better PR ;-)

PR #1186, opened 2018-05-31
black text for focus/hover on disabled dropdown

vjawala:


src/shared/styles/config/extend/_hover-focus.scss
```diff
@@ -1,7 +1,7 @@
 %active-button {
   background-color: $secondary-gray20;
-  border-color: $primary-dark;
-  color: $primary-dark;
+  border-color: $primary-dark !important;
+  color: $primary-dark !important;
   fill: $primary-dark;
```
aditudorache: important? bad practice... is not another way?

vjawala: In a few cases there is a `...__label` or something element inside the focusable element, so the !important is to make sure this style is persisted throughout

PR #1187, opened 2018-06-05
Filter layers that are not visible so user can't select layers that a…

aronmartin:
…re not visible

### DIFF
--- a/src/map/sagas/map-click/map-click.js
+++ b/src/map/sagas/map-click/map-click.js
@@ -7,10 +7,15 @@ import { getMapZoom } from '../../ducks/map/map';
 import ACTIONS from '../../../shared/actions';
 
 const getMapLayers = (state) => (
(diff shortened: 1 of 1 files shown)


src/map/sagas/map-click/map-click.js
```diff
@@ -7,12 +7,15 @@ import { getMapZoom } from '../../ducks/map/map';
...
-   layer.detailUrl && !layer.noDetail &&
-   (!layer.authScope || state.user.scopes.includes(layer.authScope))
- ))
+  state.map.overlays
+    .filter((overlay) => overlay.isVisible)
+    .map((overlay) => (
+      state.mapLayers.layers.items.find((layer) => layer.id === overlay.id) || {}
+    ))
+    .filter((layer) => (
+     layer.detailUrl && !layer.noDetail &&
```
boris-arkenaar: indenting

PR #1229, opened 2018-07-10
DP-5725 Don't reset geometry when going into drawing mode.

boris-arkenaar:
Restore previous drawing shape when exiting drawing mode without having drawn something.


boris-arkenaar:
Branch name not picked up by CI. Opening PR with another branch.

PR #1248, opened 2018-08-06
Feature/dp 5733 tekenen instabiel

boris-arkenaar:


src/map/containers/draw-tool/DrawToolContainer.jsx
```diff
@@ -95,6 +96,17 @@ class DrawToolContainer extends React.Component {
     }
   }
 
+  componentWillUnmount() {
+    this.props.destroy();
```
aditudorache: destroy is pretty general and confusing, maybe releaseReferences or something like this?

boris-arkenaar: The draw-tool service needs to destroy itself. I don't think it necessary for the draw-tool container to know what the service needs to do to destroy itself. So I would say a general name is preferable. Besides, it does more than just release references.

PR #1267, opened 2018-08-29
DP-5793 Add link from gebied to data selection for brk

boris-arkenaar:


boris-arkenaar:
Needs to be based on feature/eigendommen branch

PR #1295, opened 2018-09-17
scrolling into view click item

RRMoelker:


RRMoelker:
Doesn't resolve broken test

PR #1298, opened 2018-09-18
Feature/eigendommen

aditudorache:


aditudorache:
sync eigendommen with develop, e2e tests are fixed

PR #1307, opened 2018-09-20
Feature/dp 5861 state2url

aditudorache:


src/shared/environment.js
```diff
@@ -35,7 +35,7 @@ export const isDevelopment = () => {
 
 // This object enables/disables features that are not yet released
 export const features = {
-  eigendommen: false
+  eigendommen: true
```
chrisvanmook: Should this be enabled?

src/shared/services/routing/state-to-url.js
```diff
@@ -0,0 +1,45 @@
...
+
+    const params = stateUrlConverter.state2params(state);
+    const hash = getSanitizedHash(params);
+
+    if (!isHashChanged(location.hash, hash)) return;
+
+    // eslint-disable-next-line no-console
+    // console.log('updating url hash (location.hash and hash are different)');
+    if (useReplace) {
+      console.log('replace', hash); // eslint-disable-line no-console
```
chrisvanmook: Do we really need console.log here?

src/shared/environment.js
aditudorache: yes, we test all the existing features here, including eigendommen

src/shared/services/routing/state-to-url.js
aditudorache: removed

PR #1320, opened 2018-09-28
Feature/dp 5581 eigendommen layers

aditudorache:


public/images/map-legend/icon-egdm-cat-1.svg
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="16px" height="16px" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 47 (45396) - http://www.bohemiancoding.com/sketch -->
```
chrisvanmook: Perhaps remove the comment with title, desc and defs in all the svg's?

aditudorache: these will be regenerated from sketch when Nico is back. I'll add a reminder for that on the backlog.

PR #1328, opened 2018-10-01
DP-5572: Moved ng-if statement to upper span to prevent showing label…

royderks:
Moved ng-if statement to upper span to prevent showing label for resources with type website


royderks:
Rebased and added to PR for DP5633: https://github.com/Amsterdam/atlas/pull/1329

PR #1336, opened 2018-10-03
Feature/docker image health checks

RRMoelker:
From average pipeline stage time of about 6 minutes to 10 seconds!!! :)


RRMoelker:
Related: https://github.com/Amsterdam/atlas-health-checks

PR #1357, opened 2018-10-10
Feature/react as root homepage links

RRMoelker:


src/app/components/Footer/Footer.jsx
```diff
@@ -20,8 +20,7 @@ const FooterComponent = () => (
             >melden</a>.
           </p>
           <p className="c-dashboard__proclaimer-paragraph">
-            <Link
-              href={{ type: routing.proclaimer.type }}
+            <Link // eslint-disable-line jsx-a11y/anchor-is-valid
```
chrisvanmook: perhaps we could add an exeption for Link in the eslint config, as we will probably need to use the Link component way more in the near future

RRMoelker: I'll look into it the next time I encounter it.

PR #1404, opened 2018-10-30
Dp 6123 piwik event show all results

royderks:


modules/detail/components/detail/detail.component.test.js
```diff
@@ -617,6 +620,12 @@ describe('the dp-detail component', () => {
             const scope = component.isolateScope();
             expect(scope.vm.geosearchButton).toEqual([52.654, 4.987]);
         });
+
+        it('is handles piwik on ng-click', () => {
```
aditudorache: it handels

PR #1407, opened 2018-10-30
DP-6124: Send piwik event when user click on home logo in the header

royderks:


modules/header/components/logo/logo.component.test.js
```diff
@@ -50,6 +54,11 @@ describe('The dp-logo component', () => {
         it('shows the short logo image', () => {
             expect(component.find('.qa-logo__image')[0].getAttribute('src')).toContain('logo-short');
         });
+
+        it('is handles piwik on ng-click', () => {
```
aditudorache: it handles

PR #1410, opened 2018-10-31
Dp 6140 map highlight kadaster

royderks:


src/map/components/leaflet/__snapshots__/MapLeaflet.test.jsx.snap
```diff
@@ -2488,18 +2488,6 @@ exports[`MapLeaflet component should render map with geoJSON with RD coordinates
       opacity={0}
       url="https://acc.map.data.amsterdam.nl/maps/biz"
     />
-    <RdGeoJson
```
aditudorache: This test is not relevant for the case where points/polygon are/is rendered. Please add one more test where points/polygon are/is rendered!

PR #1436, opened 2018-11-13
Print feature working now

chrisvanmook:


modules/atlas/styles/deprecated/_dashboard-print.scss
```diff
@@ -50,3 +51,30 @@ $dashboard-heading-padding-print-version: 46px;
     border: none;
   }
 }
+
+// Todo: move this logic to js
```
RRMoelker: Hmm, I would say we have to much presentation logic in JS at the moment. I would rather have everything else moved to css. But that's way to much work. So just remove Todo?

modules/atlas/styles/deprecated/_dashboard-print.scss
```diff
@@ -50,3 +51,30 @@ $dashboard-heading-padding-print-version: 46px;
...
+// sass-lint:disable nesting-depth
+.c-dashboard--page-type {
+  &-search-data {
```
RRMoelker: Oef... Can't we just make all dashboards columns width 100%? Not just the `search-data` one?

modules/header/components/menu/_menu.scss
```diff
@@ -34,7 +34,7 @@
 
 .c-menu__item--help {
   .c-menu__title {
-    .c-dashboard--page-type-help & {
+    .c-dashboard--page-type--help & {
```
RRMoelker: These page types are a complete hack... Probably out of scope. But must be said

src/app/components/SplitScreen/SplitScreen.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/* istanbul ignore next */ // TODO: refactor, test
+const SplitScreen = ({ leftComponent, rightComponent }) => (
```
RRMoelker: Good one

src/shared/ducks/ui/ui.js
```diff
@@ -21,12 +20,16 @@ export const initialState = {
 
 export default function UiReducer(state = initialState, action) {
   switch (action.type) {
+    case routing.panorama.type:
+    case routing.dataDetail.type:
+    case routing.detail.type:
     case routing.map.type: {
-      const { embedPreview, embed } = action.meta.query || {};
+      const { embedPreview, embed, print } = action.meta.query || {};
```
RRMoelker: print from query should be read on EVERY route change I think. So if `action.type` starts with `'reduxRouter/'` then..

src/store/redux-first-router.js
```diff
@@ -32,6 +32,12 @@ export const isMapPage = createSelector(
   (currentPage) => currentPage === PAGES.KAART
 );
 
+export const isPanoPage = createSelector(
+  getCurrentPage,
+  (currentPage) => currentPage === PAGES.PANORAMA
+    || currentPage === PAGES.KAART_PANORAMA
```
RRMoelker: There should be only one Panorama "Page" right now. One of these can be trashed

modules/atlas/styles/deprecated/_dashboard-print.scss
chrisvanmook: More like those page-types selectors should be not in css... or js

modules/atlas/styles/deprecated/_dashboard-print.scss
chrisvanmook: columns has a col-sm-8 etc. classes. so width is now overridden basically, hence the Todo up above ;)

modules/header/components/menu/_menu.scss
chrisvanmook: Totally agreed!

src/shared/ducks/ui/ui.js
chrisvanmook: Good one!

src/store/redux-first-router.js
chrisvanmook: Thanks!

modules/header/components/menu/_menu.scss
chrisvanmook: Also, we should eventually get rid of the dashboard component-class...

PR #1471, opened 2018-11-26
Dp 6246 e2e fix + datasets functionality

aditudorache:


modules/page/components/catalogus-themes/catalogus-themes.component.js
```diff
@@ -19,8 +19,11 @@ import { routing } from '../../../../src/app/routes';
         vm.themes = CATALOGUS_THEMES_CONFIG.map(theme => {
             const linkTo = {
                 type: routing.datasets.type,
```
chrisvanmook: Updating query params like this is not really . a best practice. best is to dispatch an action with an action creator. Then in the filter-query.js file we can add that action.

modules/shared/components/tab-header/tab-header.factory.js
```diff
@@ -1,130 +1,15 @@
+import TabHeader from '../../../../src/shared/services/tab-header/tab-header';
+
 (function () {
     'use strict';
 
     angular
         .module('dpShared')
         .factory('TabHeader', tabHeaderFactory);
 
-    tabHeaderFactory.$inject = ['TAB_HEADER_CONFIG'];
```
chrisvanmook: 🎉 🎉 🎉 🎉 🎉

src/app/components/Dataset/Dataset.jsx
```diff
@@ -35,21 +37,24 @@ const Dataset = ({
...
+  apiSpecification,
+  searchText
 }) => {
   if (isLoading || (!numberOfRecords && !authError)) {
     return <LoadingIndicator />;
   }
 
   // Todo: move query and showTabbHeader logic to search page
-  const query = ''; // ?
+  const query = searchText || '';
```
chrisvanmook: could the empty sting be moved to the selector?

src/header/ducks/auto-suggest/auto-suggest.js
```diff
@@ -41,7 +44,13 @@ export default function AutoSuggestReducer(state = initialState, action) {
...
         ...state,
         activeSuggestion: action.suggestion,
-        displayQuery: action.suggestion.label
+        displayQuery: action.suggestion.label,
+        typedQuery: action.suggestion === -1 ? '' : state.typedQuery
+      };
+    case routing.searchDatasets.type:
+      return {
+        ...state,
+        typedQuery: get(action, 'meta.query.zoekterm')
```
chrisvanmook: default value (empty string) as 3rd argument?

src/shared/ducks/datasets/data/data.js
```diff
@@ -13,16 +15,18 @@ export const DEFAULT_VIEW = 'CATALOG';
...
-  authError: false
+  authError: false,
+  searchText: ''
 };
 
 export default function reducer(state = initialState, action) {
   switch (action.type) {
     case routing.datasets.type: {
-      const { page } = action.meta.query || {};
+      const { page, zoekterm } = action.meta.query || {};
```
chrisvanmook: I'm not 100% sure, but we have a dataSearch and I might think that is the place where you want to save the searchText in the store

src/shared/services/tab-header/tab-header.js
```diff
@@ -0,0 +1,144 @@
...
+                  optie 'Alle datasets tonen' en filter vervolgens op thema.`
+    }
+  }
+};
+
+const countProvider = {};
+const tabHeader = {};
+
+export class TabPage {
+  constructor(id, { title, action, getPayload, tip }) {
```
chrisvanmook: awwww :( for a second thought this TabHeader was removed for ever ;)

modules/page/components/catalogus-themes/catalogus-themes.component.js
aditudorache: action creator moved to redux-first-router

modules/shared/components/tab-header/tab-header.factory.js
aditudorache: :-)

src/app/components/Dataset/Dataset.jsx
aditudorache: done

src/header/ducks/auto-suggest/auto-suggest.js
aditudorache: thanks!

src/shared/services/api/api.js
```diff
@@ -1,19 +1,23 @@
...
 import SHARED_CONFIG from '../shared-config/shared-config';
 import { encodeQueryParams } from '../query-string-parser/query-string-parser';
+import { logout } from '../auth/auth';
 
 export const getAccessToken = () => getState().user.accessToken;
 
 export const generateParams = (data) => Object.entries(data).map((pair) => pair.map(encodeURIComponent).join('=')).join('&');
 
 const handleErrors = (response) => {
+  // Todo:
```
aditudorache: Ticket?

src/shared/services/api/api.js
```diff
@@ -1,19 +1,23 @@
...
+    logout();
+  }
   if (!response.ok) {
     throw Error(response.statusText);
   }
   return response;
 };
 
-
 // TODO: This function is not used yet because it is not finished
```
aditudorache: TIcket

PR #1484, opened 2018-11-29
DP-6008: Update colors for label of resourcetypes in catalogus, added new ng f…

royderks:
Update colors for label of resourcetypes in catalogus, added new ng filter to transform resourcetypes to css classes and refactor code


modules/shared/filters/remove-spaces.filter.test.js
```diff
@@ -0,0 +1,21 @@
+describe('The removeSpaces filter', function () {
```
aditudorache: reanme to kebab-case, see https://stackoverflow.com/questions/11273282/whats-the-name-for-hyphen-separated-case/12273101

royderks: Renamed without the dash, as angular doesn't support dashes in name of filters

PR #1487, opened 2018-12-05
Dp 6340 bouwblok layer 

aditudorache:


src/map/ducks/detail/constants.js
```diff
@@ -0,0 +1,4 @@
+export const FETCH_MAP_DETAIL_REQUEST = 'FETCH_MAP_DETAIL_REQUEST';
```
chrisvanmook: could you give it a namespace plz? ```export const FETCH_MAP_DETAIL_REQUEST = `${REDUCER_KEY}/FETCH_MAP_DETAIL_REQUEST`;```

chrisvanmook: however, make sure these are not used in an angular component that uses the constant string directly (what happens a lot in dp-link for example)

src/store/redux-first-router.js
```diff
@@ -25,20 +25,20 @@ export const preserveQuery = (action) => {
...
       subtype,
       id: `id${id}`
-    },
-    meta: {
-      query: {}
     }
-  };
+  });
   if (view === DETAIL_VIEW.MAP) {
     action.meta.query.kaart = '';
```
chrisvanmook: I know this might be out of scope, but could you perhaps have a look if you can use kaart=true instead of kaart=''? Makes more sense I guess, since then we can just set the kaart = true or false, then we get rid of the if statement as well

src/map/ducks/detail/constants.js
aditudorache: ok

src/store/redux-first-router.js
aditudorache: out of scope, the same is for panorama, perhaps handle them together later

PR #1498, opened 2018-12-12
DP-6259 add 404 not found to CMS

aditudorache:


src/app/pages.js
```diff
@@ -42,4 +43,5 @@ export const isCmsPage = (page) =>
   page === PAGES.AVAILABILITY_QUALITY ||
   page === PAGES.STATISTICS ||
   page === PAGES.VERPLAATST ||
+  page === PAGES.NIET_GEVONDEN ||
```
royderks: Replace with English

PR #1514, opened 2018-12-14
Feature/react as root panorama new api

royderks:
Also contains the merge with the develop branch, TODO: refactor more to delete Angular


modules/straatbeeld/straatbeeld-config.js
```diff
@@ -1,9 +1,12 @@
 export default {
-    STRAATBEELD_ENDPOINT_ALL: 'panorama/recente_opnames/alle/',
-    STRAATBEELD_ENDPOINT_YEAR: 'panorama/recente_opnames/',
+    STRAATBEELD_ENDPOINT_PREFIX: 'panorama/panoramas',
```
aditudorache: merge this definition with the one in src/shared/services/panorama-api/panorama-api.js ?

modules/detail/components/detail/templates/dcatd/_datasets.scss
```diff
@@ -17,14 +17,16 @@
     }
 
     &__right {
+      display: flex;
```
aditudorache: Does this work in IE11?

royderks: Yes https://caniuse.com/#feat=flexbox

modules/straatbeeld/straatbeeld-config.js
royderks: Fair point! This piece is no longer used and can be deleted

src/panorama/containers/PanoramaContainer.jsx
```diff
@@ -0,0 +1,91 @@
...
+const mapStateToProps = (state) => ({
+  panoramaState: getPanorama(state),
+  hotspots: getHotspots(state),
+  isPrint: isPrintMode(state),
+  // Todo: DP-6312: get the location from selection reducer
+  locationString: getMapCenter(state).join(',')
+});
+
+const mapDispatchToProps = (dispatch) => bindActionCreators({
+  doClose: toDataSearchLocation,
```
aditudorache: doClose should return to the previous view before opening the panorama, panorama is just a view mode on top of another state/route.

PR #1521, opened 2018-12-19
Feature/react as root panorama new api

royderks:


src/panorama/containers/PanoramaContainer.jsx
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  render() {
+    const { isFullscreen, panoramaState } = this.props;
+
+    return (
+      <div className="c-panorama">
+        <div
+          ref={
+            //eslint-disable-next-line
```
chrisvanmook: Perhaps also specify what should be disabled in eslint

royderks: Will update after merge // eslint-disable-next-line no-return-assign

PR #1539, opened 2019-01-02
Fix DP-6452, DP-6453: only send api calls for 'vestigingen' en 'monu…

royderks:
Fix dp-6452 + dp-6453; only send api calls for 'vestigingen' en 'monumenten' when they are present in the parent data


modules/detail/components/partial-select/partials/monumenten/adressen_lijst.html
```diff
@@ -6,10 +6,10 @@
     </h4>
     <ul class="o-list">
         <li ng-repeat="object in apiData.results">
-            <dp-redux-link
+            <!-- <dp-redux-link
```
chrisvanmook: Commented code can be removed?

PR #1551, opened 2019-01-06
Dp 6467 normalize coordinates

aditudorache:


src/map/ducks/map/map.js
```diff
@@ -48,8 +50,8 @@ export default function MapReducer(state = initialState, action) {
   const { lat, lng, zoom, legenda, lagen } = query;
   if (lat && lng) {
```
royderks: This was already in the code before, but the check for 'lat && lng' makes the ||-statement obsolute, so the enrichedState.viewCenter could become normalizeLocation([lat, lng], 7)

PR #1563, opened 2019-01-08
DP-6454 fix clear autosugest when open detail page

aditudorache:


src/header/ducks/auto-suggest/auto-suggest.js
```diff
@@ -20,6 +21,15 @@ const initialState = {
 };
 
 export default function AutoSuggestReducer(state = initialState, action) {
+  // cleanup the state for this reducer when not on the search routes
+  if (action.type &&
```
royderks: Change to function that takes ROUTER_NAMESPACE and an array of PAGES.* as we use this on more places

royderks: Or use a selector from 'store/redux-first-router'

PR #1566, opened 2019-01-09
Changed url params according to DP-6438

chrisvanmook:
Split first-router to seperate files
Added middleware to handle additional params and / or keep current params


src/shared/ducks/data-selection/constants.js
```diff
@@ -15,24 +15,34 @@ export const ROUTE_DATASET_MAPPER = {
...
-  brk: routing.cadastralObjects.type
-};
 
 export const VIEWS = {
   LIST: 'LIST',
   TABLE: 'TABLE',
   MAP: 'MAP'
 };
 
+export const CONSTANTS_PARAMS = {
```
aditudorache: naming is to general, change to something that suggests the contents of the list

src/store/params-registry/paramRegistry.js
```diff
@@ -218,16 +222,18 @@ class ParamsRegistery {
...
+   * @param parameters
    * @param route
+   * @param encode Tell if we should encode the parameters values
    * @returns {*}
    */
-  getNextQueries(additionalParams, route) {
-    return Object.entries(additionalParams).reduce((acc, [parameter, value]) => {
-      const encode = get(this.result, `${parameter}.routes[${route}].encode`);
-      return encode ? {
+  getNextQueries(parameters, route, encode = true) {
```
aditudorache: please review the function name, is not clear what it does from the name

PR #1580, opened 2019-01-15
Added piwik events for actions and routes, extended middleware, updat…

royderks:
Added piwik events for actions and routes, extended middleware, updated tests


src/store/middleware/piwik/routes.js
```diff
@@ -0,0 +1,16 @@
+import { routing } from '../../../app/routes';
+
+const mappedRoutes = {};
+// eslint-disable-next-line array-callback-return
+Object.keys(routing).map((key) => {
```
chrisvanmook: Please remove eslint-disable, I think it could be re-written like:
`
const mappedRoutes = Object.entries(routing).reduce((acc, [key, value]) => ({
  ...acc,
  [mappedRoutes[value.type] = () => [
    'trackPageView',
    routing[key].title,
    window.location.href,
    null
  ]
}), {})
`

PR #1583, opened 2019-01-16
Dp 6202 document title

aditudorache:


src/store/middleware/documentHead.js
```diff
@@ -8,7 +10,28 @@ const documentHead = () => (next) => (action) => {
   const typeIsRoute = Object.keys(routing).map((key) => routing[key].type).includes(action.type);
   if (typeIsRoute) {
     const page = Object.keys(routing).find((key) => routing[key].type === action.type);
-    const pageTitle = get(routing, `[${page}].title`, TITLE_DEFAULT);
+    const pageType = get(routing, `[${page}].type`, '');
+    let pageTitle = get(routing, `[${page}].title`, TITLE_DEFAULT);
+
+    switch (pageType) {
+      // eslint-disable-next-line no-case-declarations
```
chrisvanmook: please remove the eslint-disable, you can write cases like:
`
case routing.home.type: {
...
}
`
Also, I'm not a big fan of putting this kind of logic (logic per route) in middleware. Please consider refactoring this

aditudorache: thanks!

PR #1584, opened 2019-01-16
Feature/react as root clear cache

royderks:


src/shared/services/api/api.js
```diff
@@ -18,7 +19,25 @@ const handleErrors = (response, reloadOnUnauthorized) => {
...
+    Object.entries(sessionStorage).forEach(([itemKey]) => {
+      if (itemKey.startsWith('http')) {
+        const cache = JSON.parse(sessionStorage.getItem(itemKey));
+
+        const now = new Date().getTime();
+        let expiration = new Date(cache.timestamp);
+        expiration = expiration.setMinutes(expiration.getMinutes()
+          + SHARED_CONFIG.CACHE_EXPIRATION);
+
+        if (now >= expiration) {
```
chrisvanmook: What about if we exceed the threshold and now >= expiration === false?

royderks: True, although the threshold is fairly high we can clear the entire api cache for this powe ruser

PR #1587, opened 2019-01-16
Dp 6202 document title detail page

aditudorache:


modules/detail/components/detail/detail.component.test.js
```diff
@@ -271,7 +291,7 @@ describe('the dp-detail component', () => {
         });
     });
 
-    it('puts data on the scope based on the hr endpoint', () => {
+    xit('puts data on the scope based on the hr endpoint', () => {
```
chrisvanmook: can these xit's be removed instead?

PR #1612, opened 2019-01-23
Calculate size of sessionstorage based on the size of json objects in…

royderks:
…side is, set maximum size to 200kb


src/shared/services/api/api.js
```diff
@@ -21,7 +21,9 @@ const handleErrors = (response, reloadOnUnauthorized) => {
 
 
 const handleCache = (response, key) => {
-  if (sessionStorage.length > SHARED_CONFIG.CACHE_THRESHOLD) {
+  const sessionStorageSize = unescape(encodeURIComponent(JSON.stringify(sessionStorage))).length;
```
aditudorache: I wouldn't do this calculation, this is an expensive operation for each uncached call. Also please refactor the getFromCache function, I would return null instead of throwing an Exception, also an expensive operation that can be avoided in that situation.

PR #1614, opened 2019-01-23
Moved hacky logic to forceFullscreen for detail view

chrisvanmook:


src/shared/ducks/detail/selectors.js
```diff
@@ -12,3 +15,12 @@ export const getDetailEndpoint = createSelector(getDetail, (detail) => {
 });
 export const getDetailDisplay = createSelector(getDetail, (detail) => detail && detail.display);
 export const isDetailLoading = createSelector(getDetail, (detail) => detail && detail.isLoading);
+export const hasNoGeometry = createSelector(
```
royderks: Why not check if it has Geometry, as we can use this on more files

chrisvanmook: Instead I added a todo and changed the method name, we should actually never use this logic...

PR #1659, opened 2019-02-01
Hotfix/shape download

chrisvanmook:


src/app/components/DataSelection/DataSelection.jsx
```diff
@@ -37,6 +39,7 @@ const DataSelection = ({
   },
   page: currentPage
 }) => {
+  console.log(activeFilters);
```
aditudorache: remove console log

PR #1692, opened 2019-02-12
Dp-6673: Toggle Panorama bugs

royderks:
Cleaned up code, removed X button when panorama layer activated, preserve all maplayers except the panorama layers on toggling between views


src/header/sagas/auto-suggest/auto-suggest.js
```diff
@@ -20,17 +16,6 @@ export function* fetchSuggestions(action) {
   }
 }
 
-export function* doSelectSugestion(action) {
-  try {
-    const { suggestion, view } = action.payload;
-    const overlaysWithoutPanorama = yield select(getMapOverlaysWithoutPanorama);
-    yield put(toDataSuggestion(suggestion, view, overlaysWithoutPanorama));
-  } catch (error) {
-    yield put(setGlobalError(error.message));
```
royderks: Can I delete this here @aditudorache

aditudorache: Sure if it is not used anymore!

PR #1700, opened 2019-02-13
Feature/dp 6306 charging locations layer

royderks:


src/map/components/detail-result/__snapshots__/MapDetailResult.test.jsx.snap
```diff
@@ -341,6 +341,8 @@ exports[`MapDetailResult should render nap peilmerk 1`] = `
 />
 `;
 
+exports[`MapDetailResult should render oplaadpuntent 1`] = `""`;
```
aditudorache: this snapshot is empty, is this correct?

PR #1730, opened 2019-03-01
Dp 6426 pano viewer

royderks:


royderks:
Strange behavior with the "back"-button when switching years with the toggle button, some pages get duplicated in the history. Also see https://github.com/Amsterdam/atlas/pull/1718 that solved this issue before..

PR #1794, opened 2019-04-08
Dp 6835 enable e2e

aditudorache:


src/app/containers/DetailContainer/Detail.jsx
```diff
@@ -2,21 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { AngularWrapper } from 'react-angular';
 
+// eslint-disable-next-line react/prefer-stateless-function
```
royderks: Why not make this a function component, there are no lifecycles?

src/panorama/components/PanoramaToggle/PanoramaToggle.jsx
```diff
@@ -82,11 +82,11 @@ class PanoramaToggle extends React.Component {
                 href={getStreetViewUrl(location, heading)}
                 target="_blank"
                 rel="noopener"
-                className="
+                className={`
```
royderks: Why is this a template literal?

src/app/containers/DetailContainer/DetailContainer.jsx
```diff
@@ -28,7 +27,6 @@ const mapStateToProps = (state) => ({
 });
 
 const mapDispatchToProps = (dispatch) => bindActionCreators({
```
royderks: Whats the point of having an empty mapDispatchToProps?

src/shared/ducks/detail/actions.js
```diff
@@ -20,11 +20,6 @@ export const setView = (payload) => ({
   payload
 });
 
-export const fetchDetailRequest = (payload) => ({
```
royderks: What's the reason for deleting this pattern for Redux actions? REQUEST -> SUCCESS / FAILURE

src/panorama/components/PanoramaToggle/PanoramaToggle.jsx
aditudorache: otherwise, there is a problem with the layout in the snapshots in windows. windows adds an extra \r for each \n in the not templated strings

src/app/containers/DetailContainer/Detail.jsx
aditudorache: oke

src/app/containers/DetailContainer/DetailContainer.jsx
aditudorache: removed

src/shared/ducks/detail/actions.js
aditudorache: The fetchDetailRequest was not used anymore the SUCCESS / FAILURE are dispatched with data retrieved/transformed during the fetchMapDetail action. This was just a cleanup action, has no direct effect on this PR

PR #1795, opened 2019-04-11
Dp 6841 header splitsing

aditudorache:


src/header/services/header-size/header-size.constant.js
```diff
@@ -0,0 +1,8 @@
+const headerSize = {
```
royderks: Use uppercase for constants

src/app/components/Header/Header.test.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Header from './Header';
+
+describe('Header', () => {
```
royderks: Add test cases for all three versions of the header

src/app/App.jsx
```diff
@@ -188,7 +189,9 @@ App.propTypes = {
   printModeLandscape: PropTypes.bool.isRequired,
   embedPreviewMode: PropTypes.bool.isRequired,
   overflowScroll: PropTypes.bool.isRequired,
-  user: PropTypes.object.isRequired // eslint-disable-line react/forbid-prop-types
+  user: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
```
royderks: Write as PropTypes.shapeOf({})

src/header/services/header-size/header-size.constant.js
aditudorache: done!

src/app/components/Header/Header.test.jsx
aditudorache: sure! thanks!

src/app/App.jsx
aditudorache: fixed in the whole project!

PR #1812, opened 2019-05-09
Added InfoModalWrapper to show content from Google Sheets in Modal

chrisvanmook:


src/app/components/Modal/withModalBehaviour.jsx
```diff
@@ -25,6 +25,13 @@ function withModalBehaviour(WrappedComponent) {
       window.addEventListener(`openForm_${this.props.id}`, this.handleOpen);
     }
 
+    componentWillReceiveProps(nextProps) {
```
royderks: Unsafe and soon deprecated https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

royderks: Use componentDidUpdate() instead

PR #1835, opened 2019-05-20
Cleanup: set base font-family once 

chrisvanmook:
select fonts based on their weight and style


src/shared/styles/config/variables/_typography.scss
```diff
@@ -1,11 +1,11 @@
...
+$bold-weight: 900;
 
-$base-font-before: "Avenir LT W01 ";
+$base-font-before: "AvenirLT";
 $base-font-family-after: Arial, sans-serif;
-$base-font-family: $base-font-before + $normal-weight, $base-font-family-after;
+$base-font-family: $base-font-before, $base-font-family-after;
 
 $xs-font-size: 14px;
 $s-font-size: 16px;
```
royderks: Replace hardcoded font sizes with this variable

PR #1840, opened 2019-05-28
Added maplayers for DSM and DTM

royderks:


royderks:
Maplayers are in review bij the issuer

PR #1853, opened 2019-06-05
DP-6785 add the garbage layers to the map

aditudorache:


src/map/components/legend/MapLegend.jsx
```diff
@@ -29,6 +29,8 @@ class MapLegend extends React.Component {
       `sld_version=${MAP_CONFIG.SLD_VERSION}&`,
       `layer=${(legendItem.layers && legendItem.layers[0]) || (mapLayer.layers && mapLayer.layers[0])}&`,
       'format=image/svg%2Bxml&',
+      legendItem.imageSize ? `width=${legendItem.imageSize}&` : '',
```
royderks: Combine these two rules as they have the same conditioon

aditudorache: done

PR #1872, opened 2019-07-08
Dp 7007 pano 2019

aditudorache:


src/map/components/legend/MapLegend.jsx
```diff
@@ -29,6 +29,7 @@ class MapLegend extends React.Component {
         (mapLayer.layers && mapLayer.layers[0])}&`,
       'format=image/svg%2Bxml&',
       legendItem.imageSize ? `width=${legendItem.imageSize}&height=${legendItem.imageSize}&` : '',
```
royderks: Rewrite to use the a stringified object

src/map/services/map-layers/map-layers.config.js
```diff
@@ -75,10 +77,25 @@ const mapLayers = [
       mission_type: 'bi',
     },
   },
+  {
+    id: 'pano2019bi',
+    layers: ['panorama_new'],
+    legendItems: [{ title: '2019',
+    extraParameters: 'scale=800' }],
```
royderks: Use same notition as `params: { field: 'foo' }` to stringify in `MapLegend.jsx`

src/map/components/legend/MapLegend.jsx
aditudorache: good point!

src/map/services/map-layers/map-layers.config.js
aditudorache: done!

PR #1875, opened 2019-07-09
Use the datafetching hook to construct the info modal, added some mor…

royderks:
…e tests


royderks:
I refactored and closed you're open pull request for the info modal, as it was too far behind `develop` @chrisvanmook

PR #1879, opened 2019-07-10
Changed the logic for the saga that should only be called when the de…

royderks:
…tailview is not the map view


royderks:
This PR also includes some changes due to `prettier`

PR #1888, opened 2019-07-17
Blog

chrisvanmook:


src/app/utils/normalizeFromCMS.js
```diff
@@ -0,0 +1,53 @@
...
+  ])[0]
+  const publicationDate = new Date(pubDate)
+  const date = dateToString(publicationDate)
+  const options = {
+    year: 'numeric',
+    month: 'long',
+    day: 'numeric',
+  }
+
+  const localeDate = publicationDate.toLocaleDateString('nl-NL', options)
```
aditudorache: please use the existing formatDate service, does the same formatting

PR #1891, opened 2019-07-19
Updated the way data is retrieved from the cms

royderks:


src/app/utils/normalizeFromCMS.js
```diff
@@ -1,53 +1,61 @@
...
-    byline,
+    coverUrl,
+    included: data.included,
+    ...otherFields
   }
 }
+
+export default normalizeFromCMS
+
+// export const normalizePublicationData = (data) => {
```
chrisvanmook: what is the deal with all the commented out code? do we need it later?

royderks: Ah no, this can be deleted

PR #1914, opened 2019-07-25
Simplify cms hook

aditudorache:


src/app/utils/useFromCMS.js
```diff
@@ -1,33 +1,33 @@
...
+import React, { useEffect } from 'react'
 import { getByUrl } from '../../shared/services/api/api'
 import cmsNormalizer from '../../shared/services/cms/cms-normalizer'
 import { routing } from '../routes'
 
-function useFromCMS() {
+function useFromCMS(id, config) {
   const [results, setResults] = React.useState(null)
   const [loading, setLoading] = React.useState(false)
+  useEffect(() => {
```
chrisvanmook: I don't think the hook should use useEffect here, the component should decide for itself when to fetch the data. @royderks what's your opinion about this?

src/app/containers/ConstructionFilesContainer/ConstructionFilesContainer.test.jsx
```diff
@@ -45,7 +47,7 @@ describe('ConstructionFilesContainer', () => {
     expect(component).toMatchSnapshot()
   })
 
-  it('should get the results', () => {
+  it.only('should get the results', () => {
```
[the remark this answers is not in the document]
aditudorache: thanks!

src/app/utils/useFromCMS.js
aditudorache: we decided to keep this format for now

PR #1928, opened 2019-07-31
Added noElipsis prop to MapDetailResultItem, added these to the Parkeerzone panels

chrisvanmook:


src/map/components/detail-result/_map-detail-result.scss
```diff
@@ -123,10 +123,13 @@ $icon-height: 24px;
 
     &-value {
       min-height: 20px;
-      overflow: hidden;
       padding-right: $base-whitespace;
-      text-overflow: ellipsis;
-      white-space: nowrap;
+
+      &:not(&--no-elipsis) {
```
royderks: Double negelction feels a bid odd, why not `hasEllipsis`?

chrisvanmook: because then we have to edit the 1000+ files almost duplicate files ;)

PR #1930, opened 2019-07-31
qa- classes removed from scss, they are selectors for e2e tests

aditudorache:


src/app/components/DataSelection/DataSelection.jsx
```diff
@@ -50,7 +50,7 @@ const DataSelection = ({
 
   return (
     <div className="c-data-selection c-dashboard__content">
-      <div className="qa-data-selection-content">
+      <div className="c-data-selection-content qa-data-selection-content">
```
royderks: The removed `qa-data-selection-content` is still here? Replace with `data-test`?

src/app/components/DataSelection/__snapshots__/DataSelection.test.jsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`DataSelection should render LoadingIndicator 1`] = `
   className="c-data-selection c-dashboard__content"
 >
   <div
-    className="qa-data-selection-content"
+    className="c-data-selection-content qa-data-selection-content"
```
royderks: Same as previous

src/app/components/DataSelection/DataSelection.jsx
aditudorache: out of scope

src/app/components/DataSelection/__snapshots__/DataSelection.test.jsx.snap
aditudorache: out of scope

PR #1941, opened 2019-08-06
Changed the position of the tall header on the homepage to relative i…

royderks:
…nstead of fixed

### DIFF
--- a/src/app/components/Header/Header.jsx
+++ b/src/app/components/Header/Header.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
 import { css } from '@datapunt/asc-core'
-import { Header as HeaderComponent, styles, breakpoint, ascDefaultTheme } from '@datapunt/asc-ui'
+import { Header as HeaderComponent, styles, breakpoint } from '@datapunt/asc-ui'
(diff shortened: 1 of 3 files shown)


src/app/components/Header/Header.jsx
```diff
@@ -10,7 +10,19 @@ import HeaderMenuContainer from './HeaderMenuContainer'
 import EmbedHeader from './EmbedHeader'
 import PrintHeader from './PrintHeader'
 
-const style = css`
+const style = homePage => css`
+  ${styles.HeaderWrapperStyle} {
+    ${homePage &&
+      css`
+        @media screen and ${breakpoint('min-width', 'laptopM')({
+            theme: ascDefaultTheme,
```
aditudorache: shouldn't this get the theme implicitly? when the style is applyed. I dont' this using ascDefaultTheme explicit is a good practice.

src/app/components/Header/Header.jsx
```diff
@@ -56,7 +68,7 @@ const Header = ({
           tall={homePage}
           title="City Data"
           homeLink="/"
-          css={style}
+          css={() => style(homePage)}
```
aditudorache: can you maybe pass here the current theme?

royderks: Thanks! Can you check again?

PR #1956, opened 2019-08-12
Implemented new MenuButtons

chrisvanmook:


src/app/components/Header/HeaderMenu.jsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
-import { MenuInline, MenuToggle, MenuFlyOut, MenuItem } from '@datapunt/asc-ui'
+import { MenuInline, MenuToggle, MenuFlyOut, MenuItem, MenuButton } from '@datapunt/asc-ui'
 import { ChevronRight } from '@datapunt/asc-assets'
 import PropTypes from 'prop-types'
+import Link from 'redux-first-router-link'
```
royderks: @aditudorache recently named the import `RouterLink`, that way the `StyledLink` can become just `Link`(or `MenuLink`) as it's not really styled

PR #1977, opened 2019-08-22
Service worker poc + google sheets .finally (sentry warning) removed

chrisvanmook:


chrisvanmook:
> does this work in IE11? I don't think is supported

@aditudorache nope it doens't, but it is in Firefox on ADW and other modern browsers. Nothing will be changed in IE11. On the side though, gonna discuss to perhaps show a one time notification for IE11 users to tell them they can use a better browser on ADW :p

PR #1980, opened 2019-08-26
Refactored ArticleDetail to show downloads

chrisvanmook:


src/app/pages/ArticleDetailPage/ArticleDetailPage.jsx
```diff
@@ -60,6 +62,49 @@ const ArticleDetailPage = ({ id }) => {
   const documentTitle = title && `Artikel: ${title}`
   const linkAction = toArticleDetail(id, slug)
 
+  const buildDownloadsObject = () => {
```
aditudorache: please move this outside the page and pass the downloads object as a parameter. Maybe change name to normalizeDownloadsObject?

PR #1986, opened 2019-08-31
Use the normalized fields for the vsd templates in Angular as they do…

royderks:
…nt have their own api calls, enhanced normalization for 'monumenten' and 'vastgoed'


src/map/services/vastgoed/vastgoed.test.js
```diff
@@ -33,23 +33,6 @@ describe('The vastgoed resource', () => {
   })
 
   it('fetches the vastgoed geosearch', async () => {
```
chrisvanmook: plz remove the async keyword

PR #1989, opened 2019-09-02
Dp 7177/matomo redirected pages

royderks:


src/app/redirects.js
```diff
@@ -100,12 +100,19 @@ const resolveRedirects = () => {
...
     routePath = routesDictionary.filter(r => r.from === window.location.hash)
-    window.location.replace(routePath.length && routePath[0].to ? routePath[0].to : '/verplaatst/')
-    return true
+
+    setTimeout(() => {
+      window.location.replace(
+        routePath.length && routePath[0].to ? routePath[0].to : '/verplaatst/',
+      )
+      return true
+    }, 600) // Tries to prevent cancelling the network request to Matomo from the middleware
```
chrisvanmook: Could you describe why you chose for 600ms? Seems like an arbitrary number...

royderks: Yes it is very arbitrary, the problem is the network requests get cancelled as soon as the pathname is changed. Using a Promise and wait for it to resolve doesn't work. If any suggestions, let me know!

chrisvanmook: hmm let's look into it together, there must be a better solution :)

PR #1993, opened 2019-09-03
Dp 7094 frontpage integration

aditudorache:


aditudorache:
this branch si no longer active. the development is continued in the DP-7094-frontend branch

PR #1998, opened 2019-09-03
Dp 6862/map config test

royderks:


src/map/services/normalize/normalize.js
```diff
@@ -69,7 +69,7 @@ export const adressenPand = result => {
           : 'info'
         : false,
     isNevenadres: !result.hoofdadres,
-    year: result.oorspronkelijk_bouwjaar !== 1005 ? result.oorspronkelijk_bouwjaar : 'Onbekend', // The API returns 1005 when a year is unknown
+    year: result.oorspronkelijk_bouwjaar !== '1005' ? result.oorspronkelijk_bouwjaar : 'Onbekend', // The API returns 1005 when a year is unknown
```
aditudorache: Maybe here use a constant UNKNOWN_YEAR='1005'?

royderks: Thanks! Updated in https://github.com/Amsterdam/atlas/pull/1999

PR #2009, opened 2019-09-09
Sluggified dataset urls, redirect to 404 page if not found

chrisvanmook:


src/app/AppBody.jsx
```diff
@@ -25,6 +25,7 @@ const PublicationDetailPage = React.lazy(() => import('./pages/PublicationDetail
 const SpecialDetailPage = React.lazy(() => import('./pages/SpecialDetailPage'))
 const EditorialOverviewPage = React.lazy(() => import('./pages/EditorialOverviewPage'))
 const MapSplitPage = React.lazy(() => import('./pages/MapSplitPage'))
+const FourOhFour = React.lazy(() => import('./pages/FourOhFour'))
```
royderks: Haha, why not call this `NotFound` or `PageNotFound`?

src/app/utils/useSlug.js
```diff
@@ -0,0 +1,10 @@
+export default function useSlug(text) {
```
aditudorache: this is no real React hook, maybe toSlug would be a better name for this service?

PR #2021, opened 2019-09-12
Use ES endpoint for overview pages

chrisvanmook:


src/normalizations/cms/useNormalizedCMSResults.js
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable camelcase */
```
royderks: Can't this function be used inside the normalizer for the cms? Might become very bloated if the CMS data has to go through multiple normalizers before we're able to use it

chrisvanmook: You mean just move this function to the other normalize file? This is a normalizer for data that comes from elastic, the other normalizer is a normalizer for data that comes from the drupal api. ideally we also index the detail pages in elastic, then we can remove the other normalizer (and the dependency) and just use this one instead.

royderks: Ok!

PR #2022, opened 2019-09-13
Dp 7144/search tabs autosuggest

royderks:


src/header/containers/header-search/HeaderSearch.jsx
```diff
@@ -71,21 +77,41 @@ class HeaderSearch extends React.Component {
...
 
     if (activeSuggestion.index === -1) {
       // Load the search results
       onCleanDatasetOverview() // TODO, refactor: don't clean dataset on search
-      if (toDataset) {
-        onDatasetSearch(typedQuery)
+
+      const searchType =
+        label ||
+        // eslint-disable-next-line no-nested-ternary
```
chrisvanmook: Just a personal style thing, perhaps move the eslint-disable to the top to disable the rule for the whole file and destructure the constants?

src/header/containers/header-search/HeaderSearch.jsx
```diff
@@ -6,6 +6,12 @@ import { extractIdEndpoint } from '../../../store/redux-first-router/actions'
 import useSlug from '../../../app/utils/useSlug'
 import { VIEW_MODE } from '../../../shared/ducks/ui/ui'
 
+export const CONSTANTS = {
```
chrisvanmook: don't we use this somewhere else as well?

src/header/containers/header-search/HeaderSearch.jsx
royderks: Sure!

src/header/containers/header-search/HeaderSearch.jsx
royderks: Yes, imported them from the `app/routes.js`

PR #2033, opened 2019-09-19
Dp 7191 hide sharebar on print

aditudorache:


src/app/components/EditorialPage/EditorialPage.scss
```diff
@@ -5,9 +5,13 @@
 
   &__body {
     background-color: $primary-light;
-    min-height: calc(100% - 278px);
+    min-height: calc(100vh - 330px);
```
chrisvanmook: Where does the 330px come from? add a comment or create constant?

src/app/components/Dataset/Dataset.jsx
```diff
@@ -40,7 +41,12 @@ const Dataset = ({
   })
 
   if (numberOfRecords === 0) {
-    return <NoResultsForSearchType message={<NoDatasetsMessage />} hideLoginLink />
+    return (
+      <>
+        <NoResultsForSearchType message={<NoDatasetsMessage />} hideLoginLink />
+        {!printMode && <ShareBar topSpacing={6} />}
```
chrisvanmook: ` {!printMode && <ShareBar topSpacing={6} />}` is used on 6 places in this PR. Is it possible to create a wrapper with this in it or something?

chrisvanmook: also see that the printMode is passed a lot of time to various components, so perhaps it's really worth it to create a HOC / wrapper of some sort if possible

src/app/components/EditorialPage/EditorialPage.scss
aditudorache: Thanks, comment added.

src/app/components/Dataset/Dataset.jsx
aditudorache: Thanks! Refactored to ShareBar

PR #2040, opened 2019-09-24
Dp 7205 header footer

aditudorache:


src/app/components/Footer/ColofonLinks.jsx
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable react/no-array-index-key */
```
chrisvanmook: imho this is not a good idea, so please use unique key's here. Please read this if you haven't https://medium.com/@vraa/why-using-an-index-as-key-in-react-is-probably-a-bad-idea-7543de68b17c

src/app/components/Footer/FollowLinks.jsx
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable react/no-array-index-key */
```
chrisvanmook: same same

src/app/components/Footer/Footer.jsx
```diff
@@ -31,109 +30,24 @@ const StyledLink = ({ children, ...otherProps }) => (
...
-
-const FooterComponent = ({ ...otherProps }) => (
-  <Footer {...otherProps}>
+const Footer = ({ ...otherProps }) => (
+  <AscFooter {...otherProps}>
     <FooterTop>
       <Row>
-        <Column wrap span={{ small: 1, medium: 2, big: 4, large: 8, xLarge: 8 }}>
+        <Column wrap span={{ small: 1, medium: 2, big: 2, large: 4, xLarge: 4 }}>
           <>
```
chrisvanmook: Why the fragments? AboutBlock.jsx:127 for example doesn't use a fragment and Column has 2 children

src/app/components/Footer/HelpLinks.jsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import { FooterLinkList, FooterLinkListItem, Paragraph, Link } from '@datapunt/asc-ui'
+import styled from '@datapunt/asc-core'
+import { cmsIds, VEELGESTELDE_VRAGEN } from '../../../shared/services/cms/cms.config'
+import ActionLink from '../Links/ActionLink/ActionLink'
+import { toArticleDetail } from '../../../store/redux-first-router/actions'
+import { openFeedbackForm } from '../Modal/FeedbackModal'
+
+const veelGesteldeVragenLink = {
```
chrisvanmook: Use english where possible, this could be renamed as FAQ_LINKS ? Also you might want to consider capitalize constants :)

src/app/components/Footer/HelpLinks.jsx
```diff
@@ -0,0 +1,68 @@
...
+  slug: VEELGESTELDE_VRAGEN,
+}
+
+const ArticleLink = ({ title, id, slug }) => (
+  <ActionLink title={title} to={toArticleDetail(id, slug)} variant="with-chevron">
+    {title}
+  </ActionLink>
+)
+
+const FeedbackLink = styled(Link)`
```
chrisvanmook: This shouldn't be a link, but a Button :) Check how we use them in ArticleDetail page (I thought sth link `<Link $as="button" onClick={...} />`)

chrisvanmook: also don't forget the type="button"

src/app/components/Footer/ColofonLinks.jsx
aditudorache: thanks, replaced, there are though many other places in the code where this happens. issue filed

src/app/components/Footer/Footer.jsx
aditudorache: indeed, removed

src/app/components/Footer/HelpLinks.jsx
aditudorache: good point, thanks!

src/app/components/Footer/FollowLinks.jsx
aditudorache: done

src/app/components/Footer/HelpLinks.jsx
aditudorache: done!

PR #2041, opened 2019-09-25
Added orderAutoSuggestResults method

chrisvanmook:


src/header/services/auto-suggest/auto-suggest.js
```diff
@@ -4,6 +4,25 @@ import { getAuthHeaders } from '../../../shared/services/auth/auth'
 // Minimun length for typeahead query in backend is 3 characters
 const MIN_QUERY_LENGTH = 3
 
+/**
+ * Orders the array by the object's labels in the order defined in the method
+ * @param {Array} results
+ * @returns {*[]}
+ */
+export const orderAutoSuggestResults = results => {
+  const order = ['Data', 'Publicaties', 'Datasets', 'Artikelen']
```
royderks: In `HeaderSearch.jsx` I've used the `MAIN_PATHS` constants for this, can you check?

chrisvanmook: The constants in `MAIN_PATHS` are coupled to our front-end routes / titles. the string in the `order` are the same, however, these are labels that come from the typeahead api, and are not coupled to anything in our front-end, so I'd rather keep them unchanged. Will add a comment to it though

royderks: Not sure if we should use these in `HeaderSearch` as well then, in there the labels for the typeahead request are also used and mapped to these constants. So would either use these for this file as well, or also remove them from `HeaderSearch` and create new constants that are used in both files

PR #2048, opened 2019-09-26
Removed EditorialHeader, hide meta in article detail if field_type === 'content'

chrisvanmook:


src/app/pages/ArticleDetailPage/ArticleDetailPage.jsx
```diff
@@ -103,13 +106,16 @@ const ArticleDetailPage = ({ id }) => {
                     >
                       <Column span={{ small: 1, medium: 2, big: 4, large: 7, xLarge: 7 }}>
                         <EditorialBody>
-                          <EditorialHeader title={title}>
+                          <Heading $as="h1" gutterBottom={isContentType ? 18 : 0}>
```
royderks: Can't we replace the `Heading` with a styled one and remove the `gutterBottom` prop?

PR #2058, opened 2019-09-30
Added service worker + manifest json

chrisvanmook:


webpack.common.js
```diff
@@ -14,6 +15,15 @@ const dist = path.resolve(root, 'dist')
 function commonConfig(env) {
   const buildId = env && env.buildId ? env.buildId : env.nodeEnv
   const isDev = env.nodeEnv === 'development'
+  const isAcc = env.nodeEnv === 'acceptance'
+
+  const getApiUrl = (prefix = '') =>
```
royderks: This is already defined somewhere else, look for `API_ROOT`

royderks: Can't that logic be reused?

webpack.common.js
```diff
@@ -14,6 +15,15 @@ const dist = path.resolve(root, 'dist')
...
+    `https://${isAcc || isDev ? 'acc.' : ''}${prefix}data.amsterdam.nl/`
+
+  const apiUrl = getApiUrl('api.')
+  const cmsUrl = getApiUrl('cms.')
+
+  console.log(`${cmsUrl}(?!/jsonapi/node/notification)`)
```
royderks: ?

PR #2059, opened 2019-09-30
Dp 7293/ie fixes

royderks:


src/app/components/EditorialCard/EditorialCard.jsx
```diff
@@ -114,17 +114,28 @@ const EditorialCard = ({
         <Image src={teaserImage || notFoundImage} alt={title} square />
       </StyledCardMedia>
       <StyledCardContent>
-        <StyledCardHeading $as="h4">{shortTitle || title}</StyledCardHeading>
+        <div>
```
aditudorache: Please make a comment that this is due the flexbox problems in IE.

PR #2060, opened 2019-10-01
Sluggify the cms slug as it's not really a slug

chrisvanmook:


src/normalizations/cms/useNormalizedCMSResults.js
```diff
@@ -24,11 +24,13 @@ const normalizeObject = (data, type) => {
     ...otherFields
   } = data
 
+  const slug = useSlug(cmsSlug || title)
+
   // The type SPECIALS has a different url structure
   const to =
     type === PAGES.SPECIALS
       ? EDITORIAL_DETAIL_ACTIONS[type](uuid, field_special_type, slug)
-      : EDITORIAL_DETAIL_ACTIONS[type](uuid, slug || useSlug(title))
```
royderks: Why not use the title instead? As this we already have this value and therefore can streamline the process

PR #2076, opened 2019-10-02
Dp 7290 blur on mobile

aditudorache:


royderks:
Solution moved to https://github.com/Amsterdam/amsterdam-styled-components/pull/173, will be further developed there

PR #2081, opened 2019-10-04
Feature/dotenv

chrisvanmook:


package.json
```diff
@@ -10,17 +10,17 @@
   "scripts": {
     "build": "npm run build:prod",
```
aditudorache: build:production here

src/map/services/map.config.js
```diff
@@ -49,12 +49,6 @@ const environmentConfig = {
     },
     OVERLAY_ROOT: 'https://map.data.amsterdam.nl',
   },
-  [ENVIRONMENTS.PRE_PRODUCTION]: {
```
aditudorache: don't we have a preproduction environment anymore?

src/panorama/services/preview/panorama-preview.js
```diff
@@ -14,7 +12,8 @@ export default function fetchPano(location) {
     .map(key => `${encodeURIComponent(key)}=${encodeURIComponent(searchParams[key])}`)
     .join('&')
 
-  return getByUrl(`${SHARED_CONFIG.API_ROOT}panorama/thumbnail/?${queryString}`).then(response => ({
+  console.log(process.env.CMS_ROOT)
```
aditudorache: please remove

src/shared/environment.js
```diff
@@ -1,43 +1,10 @@
 export const ENVIRONMENTS = {
-  DEVELOPMENT: 'DEVELOPMENT',
-  ACCEPTANCE: 'ACCEPTANCE',
-  PRE_PRODUCTION: 'PRE_PRODUCTION',
```
aditudorache: why is the preproduction configuration removed?

webpack.production.js
```diff
@@ -3,16 +3,13 @@ const UglifyJSPlugin = require('uglifyjs-webpack-plugin')
...
-  const isAcc = env && env.nodeEnv === 'acceptance'
+const env = dotenv.config().parsed
 
-  // Todo: Put this in a .env file: https://datapunt.atlassian.net/browse/DP-7302
-  const getApiUrl = (prefix = '') => `https://${isAcc ? 'acc.' : ''}${prefix}data.amsterdam.nl/`
-
-  const apiUrl = getApiUrl('api.')
-  const cmsUrl = getApiUrl('cms.')
+module.exports = () => {
+  const apiUrl = env.API_ROOT
```
aditudorache: are these still used? Otherwise plz remove.

src/map/services/map.config.js
chrisvanmook: No, because using this didn't made any sense. build environments != environment variables. Pre-production's purpose is to show the app build with production settings. Introducing a pre_production environment variable in the code is somewhat an anti-pattern...

PR #2092, opened 2019-10-08
Dp 7336 remove gs

aditudorache:


src/app/App.jsx
```diff
@@ -113,7 +114,7 @@ const App = ({
   `
 
   function AppWrapper({ children }) {
-    return homePage || editorialPage ? (
+    return homePage || editorialPage || contentPage ? (
```
royderks: The condition for `hasMaxwidth` and the return is the same

modules/page/page-config.constant.js
```diff
@@ -1,15 +1,7 @@
...
     metadata: 'Actualiteit',
-    news: 'Nieuws',
-    beleid: 'Beleid',
-    help: 'Help',
-    snelwegwijs: 'Bediening portaal',
-    apis: "API's gebruiken",
-    proclaimer: 'Proclaimer',
     inloggen: 'Na inloggen',
-    statistieken: 'Statistieken',
     verplaatst: 'Pagina verplaatst',
```
royderks: Where are these still used? Couldn't find anything for the `nietgevonden` constant

src/app/pages/ContentPage.jsx
```diff
@@ -1,60 +1,36 @@
...
-            interpolateBindings={{
-              name: templateName,
-              type,
-              item,
-            }}
-          />
-          <div className="u-col-sm--offset-1 u-col-sm--7">
-            <ShareBar />
-          </div>
+const StyledRow = styled(Row)``
```
royderks: Do we need a `StyledRow` here?

aditudorache: not anymore: removed

src/app/App.jsx
aditudorache: sure! thanks!

modules/page/page-config.constant.js
aditudorache: I have checked and are not used, I've removed them

PR #2094, opened 2019-10-09
Dp 7340 h1 height

aditudorache:


src/app/components/TabBar/TabBar.jsx
```diff
@@ -1,28 +1,24 @@
...
-      )}
-      {numberOfDataResults === 0 && numberOfDatasetResults === 0 && (
-        <span className="c-tab-header__title__text">
-          Geen resultaten met &apos;
-          {searchQuery}
-          &apos;
-        </span>
-      )}
-    </h1>
+  <>
```
royderks: Would create a div around the contents of TabBar here to wrap this component, and add the margin-top from StyledHeading as a margin-bottom

aditudorache: I think you mean margin-top here. Done.

PR #2095, opened 2019-10-09
Use cms images from imgproxy

chrisvanmook:


src/app/components/EditorialCard/EditorialCard.jsx
```diff
@@ -110,43 +111,60 @@ const EditorialCard = ({
...
+  const resize = currentPage === PAGES.PUBLICATIONS ? 'fit' : 'fill'
+  const srcSet = {
+    srcSet: `${getImageFromCms(image, 100, 100, resize)} 70w,
+             ${getImageFromCms(image, 200, 100, resize)} 200w,
+             ${getImageFromCms(image, 400, 400, resize)} 400w`,
+  }
+  const sizes = {
+    sizes: `
+    (max-width: 480px) 70px,
+    (max-width: 786px) 200px,
```
aditudorache: ? is this maybe 768px?

royderks: Or check if we can import the breakpoints from the asc theme, as these are very much correlated

PR #2097, opened 2019-10-09
Bugfix/backdrop state

royderks:


src/header/components/auto-suggest/Search.jsx
```diff
@@ -11,29 +10,19 @@ const Search = ({
...
-
   React.useEffect(() => {
-    actions.setBackDrop({
-      payload: {
-        open: showBackdrop,
-        key: 'search',
-      },
-    })
-  }, [showBackdrop])
+    const setBackDropFn = val => setBackDrop(val)
```
chrisvanmook: Could you rename `val` to something more descriptive?

src/header/components/auto-suggest/Search.jsx
```diff
@@ -1,32 +1,56 @@
...
-
-    if (showSuggestions && suggestions.length) {
-      setBackDropFn(!!(showSuggestions && suggestions.length))
-    }
-  }, [showSuggestions, suggestions.length])
-
   return (
     <React.Fragment>
-      <SearchBar showAt="tabletM" inputProps={inputProps} {...searchBarProps} showSuggestions />
+      <StyledSearchBar
```
chrisvanmook: perhaps an aria attribute as there're mouseevents?

PR #2115, opened 2019-10-30
Remove windows OS from travis.

ebabel-eu:


.travis.yml
```diff
@@ -2,7 +2,6 @@ sudo: false
 os:
   - linux
   - osx
```
chrisvanmook: might as well remove osx :)

PR #2116, opened 2019-10-30
SearchPage

chrisvanmook:


graphql/package.json
```diff
@@ -1,13 +1,14 @@
 {
   "name": "dataportaal-graphql",
-  "version": "1.0.0",
-  "main": "src/index.js",
   "license": "MPL-2.0",
   "author": "Datapunt Amsterdam",
+  "version": "1.0.0",
+  "main": "src/index.ts",
   "scripts": {
+    "generate": "graphql-codegen --config codegen.yml",
```
royderks: Shouldn't we keep the types up to date at all times, by including this in the start or build scripts?

graphql/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import dataSearchResolvers from './data-search/resolvers'
+import dataSearchSchema from './data-search/schema'
+
+const app = express()
+
+app.use(cors())
+
+const schema = buildSchema(`
+  type Query {
+    dataSearch(q: String!, limit: Int, types: [String!]): DataSearchResult
```
royderks: Would create an Input type in the schema for this

src/app/pages/SearchPage/SearchPage.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+    ${({ showFilter }) =>
+      showFilter &&
+      css`
+        display: block;
+      `}
+  }
+`
+
+const BackDrop = ({ children, element }) => ReactDOM.createPortal(children, element)
```
royderks: We should check again if the PR in `asc-ui` can't be used to handle this, as this feels like duplicate code https://github.com/Amsterdam/amsterdam-styled-components/pull/173

src/app/pages/SearchPage/SearchPage.jsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  // Enable / disable body lock when opening the filter on mobile
+  React.useEffect(() => {
+    const action = showFilter ? disableBodyScroll : enableBodyScroll
+    action(filterRef.current)
+  }, [filterRef, showFilter])
+
+  return (
+    <SearchContext.Provider
```
royderks: If we wrap this component with someting like `urql` we already have a context in which the results for the GraphQL query is available, including extra features like a refetch. Also, we don't need to send the query using a custom Hook to the server that way.

PR #2121, opened 2019-11-08
Splitted the special pages into different types, removed unused code

royderks:


src/app/pages/SpecialDetailPage/specials/Dashboard.jsx
```diff
@@ -0,0 +1,68 @@
...
+import setIframeSize from '../../../../shared/services/set-iframe-size/setIframeSize'
+import LoadingIndicator from '../../../../shared/components/loading-indicator/LoadingIndicator'
+
+const IFrameContainer = styled.div`
+  position: relative;
+  text-align: center;
+  width: 100%;
+
+  & iframe {
+    min-height: 80vh;
```
chrisvanmook: why the 80vh?

src/app/components/Video/__snapshots__/Video.test.jsx.snap
```diff
@@ -3,6 +3,7 @@
 exports[`Video should render 1`] = `
```
ebabel-eu: Is that good to include snapshots into the repository?

src/app/pages/SpecialDetailPage/specials/Dashboard.jsx
```diff
@@ -0,0 +1,68 @@
...
+    setIframeLoading(false)
+
+    // Handle resize after the iframe is loaded
+    handleResize(setIframeHeight)
+  }
+
+  React.useEffect(() => {
+    window.addEventListener('resize', handleResize)
+
+    return function cleanup() {
```
chrisvanmook: why not just `return () => `?

src/app/components/Video/__snapshots__/Video.test.jsx.snap
royderks: For new components we don't use snapshots any more, this however is an old component

PR #2122, opened 2019-11-11
Make sure autosuggest is displayed on mobile, remove blurred backgrou…

royderks:
…nd for the infomodal, change behevior for onclick events on the menu


src/header/components/auto-suggest/_auto-suggest.scss
```diff
@@ -115,11 +115,15 @@ $clear-button-margin: 3px;
...
     position: absolute;
     right: $site-header-search-form-height + $base-whitespace * 0.5;
-    top: $site-header-search-form-height - 3px;
+
+    // Makes sure the autosuggest box starts behind the SearchBar
+    top: $site-header-search-form-height - 5px;
 
     @media (max-width: 768px) {
-      top: 120px;
+      top: 70px;
```
chrisvanmook: perhaps a comment why 70px?

PR #2131, opened 2019-11-14
First draft for replacing the legacy search with GraphQL

royderks:


src/app/utils/usePagination.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import { useQuery } from 'urql'
+
+const usePreviousValue = track => {
```
chrisvanmook: could you rename the method and parameter to sth more verbose?

src/app/utils/usePagination.js
```diff
@@ -0,0 +1,46 @@
...
+  const ref = React.useRef(track)
+  React.useEffect(() => {
+    ref.current = track
+  })
+  return ref.current
+}
+
+const usePagination = (query, input, limit = 50, initialFrom) => {
```
chrisvanmook: can you rename initialFrom and give it a default value (or switch limit and initialFrom)?

src/app/utils/usePagination.js
```diff
@@ -0,0 +1,46 @@
...
+  const [result, setResult] = React.useState({})
+  const [from, setFrom] = React.useState(initialFrom)
```
chrisvanmook: what is from?

royderks: It's the offset, so you have `from` and `limit` like this `from=0&limit=50` = first fifty

PR #2134, opened 2019-11-20
Dp 7209/bag updates

royderks:


modules/detail/components/detail/detail.component.test.js
```diff
@@ -87,54 +87,7 @@ describe('the dp-detail component', () => {
             return q.promise
           },
         },
-        endpointParser: {
```
chrisvanmook: hahaha holy shit... glad you removed this 👍

src/detail/services/data-formatter/data-formatter.js
```diff
@@ -34,13 +34,12 @@ export const formatData = (data, subject, catalogFilters) => {
   switch (subject) {
```
chrisvanmook: meh, can we just use an if-else statement here?

PR #2137, opened 2019-11-22
No longer configure the type of cms content in the config, but get th…

royderks:
…is type from the cms. Added missing test file for the cms normalizer

### DIFF
--- a/src/app/components/HomePage/SpecialCard.jsx
+++ b/src/app/components/HomePage/SpecialCard.jsx
@@ -122,7 +122,7 @@ SpecialCard.defaultProps = {
   teaser: '',
   intro: '',
   teaserImage: '',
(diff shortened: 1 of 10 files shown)


src/app/components/HomePage/SpecialCard.jsx
```diff
@@ -122,7 +122,7 @@ SpecialCard.defaultProps = {
   teaser: '',
   intro: '',
   teaserImage: '',
-  to: null,
+  to: {},
```
chrisvanmook: doesn't this throw a warning from redux-first-router?

royderks: Haven't seen one, as the `null` was thowing an error

PR #2158, opened 2019-12-17
Deleted the grex code from the repo as this data is no longer maintained

royderks:


src/shared/services/auth/auth.js
```diff
@@ -43,7 +43,7 @@ export const SCOPES = {
   'MON/RDM': 'MON/RDM',
   'HR/R': 'HR/R',
   'BD/R': 'BD/R',
-  'GREX/R': 'GREX/R',
+  'BD/X': 'BD/X',
```
chrisvanmook: what's bd/x?

PR #2170, opened 2020-01-06
Search mobile filters

chrisvanmook:


src/app/components/Header/Header.jsx
```diff
@@ -9,6 +9,8 @@ import HeaderMenuContainer from './HeaderMenuContainer'
 import EmbedHeader from './EmbedHeader'
 import PrintHeader from './PrintHeader'
 
+export const HEADER_HEIGHT_SMALL = 50
```
royderks: I think we export a constants variable from asc-ui

PR #2186, opened 2020-01-21
DI-167 Add a basic loading skeleton for the search page

jonkoops:


src/app/pages/SearchPage/SearchResults.jsx
```diff
@@ -1,50 +1,66 @@
...
+import DatasetSearchResults from './DatasetSearchResults'
 
-const ResultItem = styled.div`
-  margin-bottom: ${themeSpacing(18)};
-`
+const MAX_ITEMS = 5
+
+const SearchResults = ({ page, ...props }) => {
+  if (EDITORIAL_SEARCH_PAGES.includes(page)) {
+    const { label, type } = SEARCH_PAGE_CONFIG[page]
```
royderks: Can't you move this up, as you also use the label for the results components of the other search pages in this component

royderks: Maybe even define which component should be rendered, and spread the props on that component as they take same the props:



Something like this:

```js
const Results = (EDITORIAL_SEARCH_PAGES.includes(page)) ? EditorialResults : ComponentXXX
<Results
        data-test={page}
        {...{
          ...props,
          label,
          type,
        }}
      />
```

src/app/pages/SearchPage/SearchPageResults.jsx
```diff
@@ -157,4 +158,30 @@ const SearchPageResults = ({
...
+  query,
+  sort,
+  allResultsPageActive,
+}) => {
+  const setTitle = label => (isOverviewPage ? label : `${label} met '${query}'`)
+
+  return (
+    <>
+      <StyledHeading>{setTitle(SEARCH_PAGE_CONFIG[currentPage].label)}</StyledHeading>
+      <FilterWrapper>
```
royderks: Do we already want users to filter when the data is loading? Otherwise I would move the render of the SearchPageResultsSkeleton down to where the SearchResultsOverview is rendered. Otherwise the Filter might appear/disappear for a split second as it's defined twice

jonkoops: Yeah, this is a less than ideal situation. I've updated the PR so that `SearchPageResultsSkeleton` no longer exists separately, instead opting to use a single component with conditions based on the loading state.

src/app/pages/SearchPage/SearchResults.jsx
jonkoops: Yeah that's an option. I haven't actually changed anything in this component but for some reason Github makes a mess out off the diff.

PR #2191, opened 2020-01-22
Fix the query for dataset filters

jonkoops:


jonkoops:
@chrisvanmook This will have to be added to the release.

PR #2194, opened 2020-02-03
Refactor search filter and add a dropdown filter box

jonkoops:


src/app/components/SearchFilter/filters/RadioFilter.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import { RadioGroup, Radio, Label } from '@datapunt/asc-ui'
+import { FilterProps } from '../models'
+import { getOptionValue, formatOptionLabel, formatAllOptionLabel } from '../utils'
+
+const RadioFilter: React.FC<FilterProps> = ({ type, options, totalCount, hideCount, selection, onSelectionChange }) => {
+  function onChange(event: React.FormEvent<HTMLInputElement>) {
+    const { value: changedValue } = event.target as HTMLInputElement
+
+    if (changedValue === '') {
```
royderks: Can be written shorter

src/app/components/SearchFilter/filters/SelectFilter.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import { Select } from '@datapunt/asc-ui'
+import { FilterProps } from '../models'
+import { getOptionValue, formatOptionLabel, formatAllOptionLabel } from '../utils'
+
+const SelectFilter: React.FC<FilterProps> = ({ type, options, totalCount, hideCount, selection, onSelectionChange }) => {
+  function onChange(event: React.FormEvent<HTMLSelectElement>) {
+    const { value: changedValue } = event.target as HTMLSelectElement
+
+    if (changedValue === '') {
```
royderks: Can also be written shorter

src/app/components/SearchFilter/filters/SelectFilter.tsx
```diff
@@ -0,0 +1,38 @@
...
+      onSelectionChange([])
+    } else {
+      onSelectionChange([changedValue])
+    }
+
+    onSelectionChange([changedValue])
```
royderks: Why is this function called twice?

src/app/components/SearchFilter/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { FilterOption } from '../../models'
+
+export function formatOptionLabel({ count, label }: FilterOption, hideCount: boolean) {
+  return count !== 0 && !hideCount ? `${label} (${count.toLocaleString('nl-NL')})` : label
```
royderks: Can we define the value of `toLocaleString` somewhere in a constant?

src/app/models/index.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface Filter {
+  type: string;
+  label: string
+  filterType: FilterType
+  options: FilterOption[]
+}
+
+
+export interface FilterOption {
+  id: string
```
royderks: Can you check if we still need to have both `id` and `enumType`?

src/app/utils/usePagination.js
```diff
@@ -112,11 +112,9 @@ const usePagination = (config, input, sortString, limit, initialFrom) => {
     variables: { limit, from, sort, ...input },
   })
 
-  // Reset the state when changing the query, searchQuery or filters
-  // Todo: consider unify types and filters
   useLayoutEffect(() => {
     dispatch({ type: RESET, payload: initialState })
-  }, [config.query, input.q, input.types, input.filters])
```
royderks: Doesn't this break the interaction when switching between types of data, like between `straatnamen` and `meetbouten`

src/app/components/SearchFilter/filters/SelectFilter.tsx
jonkoops: Left it in there by accident, removed.

src/app/components/SearchFilter/filters/SelectFilter.tsx
jonkoops: If you are referring to using `if (!changedValue)` then yes, it could be written shorter. It's a matter of preference, I like to be explicit about what it is being compared to.

src/app/components/SearchFilter/utils.ts
jonkoops: Yeah, I agree this should be a application wide constant. Changed this in all places where we are using the same locale string.

src/app/models/index.ts
jonkoops: Yes, I'll take a look to see if we can get rid of this.

src/app/utils/usePagination.js
jonkoops: Yeah, this is breaking right now. But it's breaking regardless of whether `input.types` is present or not. I'm going to investigate further.

src/app/components/DocumentCover/DocumentCover.jsx
```diff
@@ -37,7 +37,7 @@ const StyledButton = styled(Button)`
   justify-content: center;
 `
 
-const DocumentCover = ({ imageSrc, onClick, title, description, loading, ...otherProps }) => (
+const DocumentCover = ({ imageSrc, onClick, description, loading, ...otherProps }) => (
   <DocumentCoverStyle {...otherProps}>
     <DocumentCoverContentStyle>
       <StyledImage src={imageSrc} alt="" />
```
chrisvanmook: Just noticed, we could use the title on the alt attribute here

src/app/components/SearchFilter/filters/CheckboxFilter.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const CheckboxFilter: React.FC<FilterProps> = ({
+  type,
+  options,
+  hideCount,
+  selection,
+  onSelectionChange,
+}) => {
+  function onChange(event: React.FormEvent<HTMLInputElement>) {
+    const { value: changedValue } = event.target as HTMLInputElement
```
chrisvanmook: shouldn't event.target already be typed by `React.FormEvent<HTMLInputElement>` ?

src/app/pages/SearchPage/SearchPage.jsx
```diff
@@ -174,29 +172,15 @@ const SearchPage = ({ isOverviewPage, currentPage, query }) => {
   }, [showFilter, currentPage])
 
   useEffect(() => {
```
chrisvanmook: yay! looking way better!

src/app/pages/SearchPage/SearchPageDucks.ts
```diff
@@ -0,0 +1,96 @@
...
+        sort: action.payload,
+      }
+    }
+
+    default:
+      return enrichedState
+  }
+}
+
+export const getQuery = ({ [REDUCER_KEY]: { query } }: { [REDUCER_KEY]: SearchPageState }) =>
```
chrisvanmook: Can't we create a type that we can use for `{ [REDUCER_KEY]: SearchPageState }`?

src/app/utils/typedAction.ts
```diff
@@ -0,0 +1,41 @@
+export function typedAction<T extends string>(type: T): { type: T }
+export function typedAction<T extends string, P extends any>(
+  type: T,
+  payload: P,
+): { type: T; payload: P }
+
+/**
+ * Creates a Redux action that is properly typed so it's payload can be inferred based on it's action type.
```
chrisvanmook: Nice!

src/app/components/DocumentCover/DocumentCover.jsx
jonkoops: Good point, fixed.

src/app/components/SearchFilter/filters/CheckboxFilter.tsx
jonkoops: Yeah, this should be `currentTarget` instead of `target`. Since `target` is not always the element that the event is attached to. Fixed.

src/app/pages/SearchPage/SearchPageDucks.ts
jonkoops: Yeah, should be possible. Nice idea! Added it :)

src/app/utils/usePagination.js
jonkoops: Ok, turns our we had the filter name hardcoded to the wrong value. It seems to work as expected now.

src/app/models/index.ts
jonkoops: Looks like we would have to change the `cms_search` implementation for this to work. It looks like this occurs mostly for filters on datasets.

jonkoops: Fixed.

PR #2197, opened 2020-02-05
Feature/css filterbox fix

chrisvanmook:


jonkoops:
Looks like this is still breaking in unexpected ways on some devices.

**iPad (landscape)**
This happens if you open the filter menu while in portrait, and then rotate the device.
image

PR #2203, opened 2020-02-10
Replace editorial and dataset overview pages with search pages

jonkoops:


src/app/pages/SearchPage/SearchPage.jsx
```diff
@@ -29,22 +25,31 @@ function getSortIntput(sortString) {
...
 
   const { documentTitle } = useDocumentTitle()
   const { trackPageView } = useMatomo()
 
-  const isSearchPage = [...EDITORIAL_SEARCH_PAGES, ...DATASET_SEARCH_PAGES].includes(currentPage)
-  const isDataSearchPage = DATA_SEARCH_PAGES.includes(currentPage)
-
   // Pagination is needed on the search pages with a single query unless the dataSearchQuery which also needs activeFilters
-  const withPagination = isSearchPage || !!(isDataSearchPage && activeFilters.length > 0) || false
+  const withPagination = isSearchPage || (isDataSearchPage && activeFilters.length > 0) || false
```
royderks: `isOverviewPage` and/or `isSearchPage` should be simplified, so it can be written like this:

`  const withPagination =
    (!isDataSearchPage && !isOverviewPage) ||
    (isDataSearchPage && activeFilters.length > 0) ||
    false`

src/app/redirects.js
```diff
@@ -102,12 +102,26 @@ const shortUrls = [
...
+}))
+
+const overviewPaths = [
+  MAIN_PATHS.ARTICLES,
+  MAIN_PATHS.PUBLICATIONS,
+  MAIN_PATHS.SPECIALS,
+  MAIN_PATHS.DATASETS,
+]
+const overviewUrls = overviewPaths.map(pathName => ({
+  from: `/${pathName}/`,
```
royderks: Check if we can also redirect if `/${pathName}` without the last slash

jonkoops: I've changed the code to handle redirects to normalize the URL before matching them. This should take care of any trailing slashes.

src/app/routes.js
```diff
@@ -49,7 +43,7 @@ export const routing = {
     page: PAGES.CADASTRAL_OBJECTS,
   },
   datasetSearch: {
-    title: 'Datasets zoekresultaten',
+    title: 'Datasets',
```
chrisvanmook: How do we set the right document title when the client is on the search page?

src/app/components/NoSearchResults/utils.ts
```diff
@@ -0,0 +1,12 @@
+// eslint-disable-next-line import/prefer-default-export
```
chrisvanmook: can use a export default here, right?

jonkoops: Yes, but I'd rather not since in the future more methods might be added here.

src/app/pages/SearchPage/SearchPage.jsx
jonkoops: I've tried the logic above, but it removes the pagination on the pages if you do not have a search query. I've added some documentation to clarify what is happening.

As far as `isOverviewPage` is concerned, I would like to remove it in the future but consider that out of scope for this PR.

src/app/routes.js
jonkoops: Since Matomo also tracks the URL and it's query parameters it is possible to keep tracking this information.

PR #2232, opened 2020-02-21
Fix issues causing ES2015+ code ending up in production build

jonkoops:
- Switch from various Babel configs to @babel/preset-env
- Automatically include core-js polyfills based on usage (useBuiltIns: 'usage')
- Pass packages written using ES2015+ syntax through Babel for transpilation to ES5
- Move browserlist config to package file
- Expand polyfills chunk with core-js and isomorphic-fetch
- Remove various dependencies that are no longer needed


babel.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        useBuiltIns: 'usage',
+        corejs: 3,
+        modules: 'commonjs',
```
jonkoops: This line actually makes our app bundle ~1mb bigger, but removing it causes errors when running the app. This needs to be investigated and resolved before we merge this PR.

src/test-index.js
```diff
@@ -1,7 +1,3 @@
-/* eslint-disable import/first */
-/* eslint-disable import/no-extraneous-dependencies */
-import 'core-js'
```
chrisvanmook: can't we move this to the webpack entry?

jonkoops: We could but it's no longer needed as `@babel/preset-env` will automatically scan our source files and add the appropriate imports to `core-js` where needed.

For more info on this check: https://babeljs.io/docs/en/babel-preset-env#usebuiltins-usage. Or run the build with [debug enabled](https://babeljs.io/docs/en/babel-preset-env/#debug) to see it in action.

babel.config.js
chrisvanmook: Does this also mean that the es directory that we ship with ASC packages is basically useless?

jonkoops: No, it's not useless. It's just that Babel + WebPack cannot deal with the transpilation properly. It looks like Babel isn't handing the modules over to WebPack for bundling but I am not sure why.

jonkoops:
Closing this PR in favor of a new one that will hopefully fix our build issues.

PR #2251, opened 2020-02-26
Add error handling for dataset and editorial search results

jonkoops:


src/app/components/EditorialResults/EditorialResults.jsx
```diff
@@ -21,24 +24,43 @@ const EditorialCardContainer = styled(CardContainer)`
...
-      </>
-    )}
-  </EditorialCardContainer>
-)
+const EditorialResults = ({ query, results, errors, label, loading, type, className }) => {
+  const matchingErrors = getErrorsForPath(errors, ['articleSearch'])
+  const hasLoadingError = matchingErrors.some(err => err.extensions?.code !== 'UNAUTHORIZED')
+
+  // Get all the labels of the type that the user has no access to
+  const unauthorizedLabels = matchingErrors.reduce(
```
royderks: Shouldn't this be a generic function? See other `*Results.jsx` files

jonkoops: Yes, I also think we can extract this. I'll get on it.

src/app/pages/SearchPage/DataSearchResults.tsx
```diff
@@ -1,37 +1,54 @@
...
   const totalCount = results.length ? [0, ...results].reduce((acc, cur) => acc + cur.count) : 0
 
-  // Get all the labels of the type that the user has no access to
-  const unauthorized =
-    errors.length &&
-    errors
-      .filter(({ code }) => code && code === 'UNAUTHORIZED')
-      .map(({ label }) => label && label.toLowerCase())
+  // Get the errors and labels for this page.
+  const matchingErrors = getErrorsForPath(errors, ['dataSearch'])
```
royderks: How about one function that returns different types of Errors? Lines 40-44 are executed in three different components, having one function that returns the loading and unauthorized error could simplify this

src/app/pages/SearchPage/SearchResultsOverview.jsx
```diff
@@ -28,23 +29,14 @@ const SearchResultsOverview = ({ query, totalCount, results, errors, loading })
...
-              // Todo: add code to error in cms_search
-              message === 'The entered type(s) does not exist',
-          )
-
+        const matchingErrors = getErrorsForPath(errors, [
+          SEARCH_PAGE_CONFIG[resultItemType].resolver,
+        ])
+        const hasLoadingError = matchingErrors.some(err => err.extensions?.code !== 'UNAUTHORIZED')
+        // TODO: This statement should match the error code instead, needs to be implemented in the API.
+        const hasNoMatchingFilters = matchingErrors.some(
```
royderks: Do we want to display this specific error or just the errors that are either unauthorized or everything else? If so, we don't have to look for this specific error. The `The entered type(s) does not exist` only occurs when a filter is used that isn't returned by the filters object - and couldn't occur

src/app/pages/SearchPage/DataSearchResults.tsx
jonkoops: Yeah. honestly I can see this happening in the future. Especially the following line seems error prone and I would like to avoid duplicating it:

```ts
  const loadingErrors = matchingErrors.filter(err => err.extensions?.code !== 'UNAUTHORIZED')
```

jonkoops: Ok, I've added a new method for getting loading errors. I think writing some generic error handling would be nice in the future, but I am considering it out of scope for now.

src/app/pages/SearchPage/SearchResultsOverview.jsx
jonkoops: True, I've changed the code to use the broader `errors` here.

PR #2252, opened 2020-02-26
Reset the page when an user filters or sorts, to always see the top r…

royderks:
…esults

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4300,104 +4300,6 @@
         }
       }
     },
-    "@storybook/addons": {
-      "version": "5.3.13",
-      "resolved": "https://registry.npmjs.org/@storybook/addons/-/addons-5.3.13.tgz",
-      "integrity": "sha512-Yjn8PXhSYw4bnB/l8ybhMw3eqpJlV3HlwMECYhRtVm+wTqal9OQlZkam5xclWm6PF/6DhyiPCt6vlXN3vfVGPA==",
-      "dev": true,
-      "requires": {
-        "@storybook/api": "5.3.13",
-        "@storybook/channels": "5.3.13",
-        "@storybook/client-logger": "5.3.13",
-        "@storybook/core-events": "5.3.13",
-        "core-js": "^3.0.1",
-        "global": "^4.3.2",
-        "util-deprecate": "^1.0.2"
-      }
-    },
-    "@storybook/api": {
-      "version": "5.3.13",
-      "resolved": "https://registry.npmjs.org/@storybook/api/-/api-5.3.13.tgz",
-      "integrity": "sha512-ZOBUOz0aRPD7w8UngcVOjwcXt417Zboi0ERW+Atn22ZUYqI3zri8M9EerLHvyKw7J42/gEIqQqg5PyfPcLJmeQ==",
-      "dev": true,
-      "requires": {
-        "@reach/router": "^1.2.1",
-        "@storybook/channels": "5.3.13",
-        "@storybook/client-logger": "5.3.13",
-        "@storybook/core-events": "5.3.13",
-        "@storybook/csf": "0.0.1",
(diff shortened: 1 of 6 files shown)


src/app/pages/SearchPage/SearchSort.jsx
```diff
@@ -53,13 +54,26 @@ const SearchSort = ({ sort, isOverviewPage, disabled }) => {
...
+            dispatch(setPage(1))
           }}
         >
-          {!isOverviewPage && <option value="">Relevantie</option>}
-          <option value={!isOverviewPage ? 'date:desc' : ''}>Publicatiedatum aflopend</option>
-          <option value="date:asc">Publicatiedatum oplopend</option>
-          <option value="title:asc">Titel A-Z</option>
-          <option value="title:desc">Titel Z-A</option>
+          {!isOverviewPage && (
+            <option key={1} value="">
```
jonkoops: Do we need to add the `key` prop to these `option` elements? I haven't seen this anywhere else before.

src/app/pages/SearchPage/SearchSort.test.jsx
```diff
@@ -0,0 +1,38 @@
+import { cleanup, fireEvent, render } from '@testing-library/react'
+import * as React from 'react'
```
jonkoops: We have have synthetic imports enabled so we could write this as:

```ts
import React from 'react'
```

However I am seeing this in other places in the project, so not that important.

package-lock.json
```diff
@@ -4300,104 +4300,6 @@
         }
```
jonkoops: I see that lockfile has been updated, but there are no change to the package file itself? :thinking:

src/app/pages/SearchPage/SearchSort.test.jsx
```diff
@@ -0,0 +1,38 @@
+import { cleanup, fireEvent, render } from '@testing-library/react'
```
jonkoops: Since this is a newly introduced file let's use the TypeScript extension (`.tsx`) for it.

src/app/components/SearchFilter/SearchFilter.tsx
```diff
@@ -48,6 +48,7 @@ const SearchFilter: React.FC<SearchFilterProps> = ({ filter, totalCount, hideCou
       trackEvent({ category: 'search', action: 'enable-filter', name: `${type}-${value}` }),
     )
 
+    dispatch(setPage(1))
```
jonkoops: Instead of manually setting the page number to the first page maybe we can just change the page number back to it's original state in the `setFilterValues` and `setSort` actions. Or are there be scenarios where these change but we don't want to change the page number? I can't come up with any.

src/app/pages/SearchPage/SearchSort.jsx
royderks: No, this can be deleted

package-lock.json
royderks: This got created when I checkout out `develop`

src/app/components/SearchFilter/SearchFilter.tsx
royderks: Hmm not a fan of handling side effects in actions. Otherwise we would have to create a composition model for actions, something we don't do at this moment and can do for a lot of other actions. I think it's more readable and easier to debug if we keep it like this, as there are only two actions dispatched

package-lock.json
royderks: I'm probably using a different version of node

jonkoops: Ok, no harm done I guess :man_shrugging:

src/app/components/SearchFilter/SearchFilter.tsx
jonkoops: Ok, then let's keep it like this.

PR #2269, opened 2020-03-02
Ensure that errors in results always have a value

jonkoops:


src/app/components/EditorialResults/EditorialResults.jsx
```diff
@@ -27,6 +27,9 @@ const EditorialCardContainer = styled(CardContainer)`
 `
 
 const EditorialResults = ({ query, results, errors, label, loading, type, className }) => {
+  // eslint-disable-next-line no-param-reassign
+  errors = errors ?? []
```
chrisvanmook: is there a way we can assign this on a higher level? so we don't need to repeat this is several files?

jonkoops: Yes. I've cleaned this up

PR #2273, opened 2020-03-02
Feature/cardlist dossiers

royderks:


src/map/services/normalize/normalize.test.js
```diff
@@ -280,7 +280,7 @@ ${input.gebruiksdoel[1]}`,
       output = kadastraalObject(input)
 
       expect(output).toMatchObject({
-        size: '1.121 m²',
+        size: '1,121 m²',
```
jonkoops: Should this not be a `.`?

src/app/components/CardList/CardList.tsx
```diff
@@ -0,0 +1,66 @@
...
+  React.useEffect(() => {
+    ;(async () => {
+      await fetchData()
+    })()
+  }, [])
+
+  return (
+    <StyledCard isLoading={loading}>
+      <StyledCardContent>
+        <StyledHeading $as="h4" styleAs="h3">
```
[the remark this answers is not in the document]
royderks: Chris will solve this later, as for now using `forwardedAs` gives type errors

src/app/components/CardList/CardListBlock.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Row, Column, themeColor } from '@datapunt/asc-ui'
+import styled from '@datapunt/asc-core'
+import CardList from './CardList'
+
+const StyledRow = styled(Row)`
+  background-color: ${themeColor('tint', 'level3')};
+`
+
+const CardListBlock = () => {
+  const lists = ['List 1', 'Second list', 'Another List', 'List 4']
```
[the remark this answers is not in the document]
royderks: Depends, we'll need the shape of the test data first. But this can be added during implementation

src/app/utils/useFromCMS.ts
```diff
@@ -3,12 +3,37 @@ import { getByUrl } from '../../shared/services/api/api'
...
+  title: string
+  shortTitle?: string
+  teaser: string
+  linkProps: Object
+  teaserImage?: string
+  coverImage?: string
+}
+
+function useFromCMS(config: CMSConfig, id = false, normalizeFromJSONApi = true): CMSResults {
+  const [results, setResults] = React.useState([])
```
jonkoops: You can use a generic here to specify the type of this hook:

```ts
const [results, setResults] = React.useState<CMSResultItem[]>([])
```

src/app/utils/useFromCMS.ts
```diff
@@ -3,12 +3,37 @@ import { getByUrl } from '../../shared/services/api/api'
...
   const [loading, setLoading] = React.useState(true)
   const [error, setError] = React.useState(false)
 
-  const fetchData = async endpoint => {
+  const fetchData = async (endpoint: string): Promise<Array<Object>> => {
```
jonkoops: This type definition can be inferred to `Promise<CMSResultItem[]>` if `results` is typed.

src/app/components/CardList/CardList.tsx
```diff
@@ -0,0 +1,73 @@
...
+  margin: ${themeSpacing(3, 0, 6)};
+`
+
+type CardListProps = {
+  title: string
+  list: CMSConfig
+}
+
+const CardList: React.FC<CardListProps> = ({ title, list }) => {
+  const { results, fetchData, loading, error } = useFromCMS(list, undefined)
```
jonkoops: There is no need to pass the second parameter here as it has a default.

src/app/utils/useFromCMS.ts
```diff
@@ -3,12 +3,37 @@ import { getByUrl } from '../../shared/services/api/api'
...
+  specialType: string | null
+  title: string
+  shortTitle?: string
+  teaser: string
+  linkProps: Object
+  teaserImage?: string
+  coverImage?: string
+}
+
+function useFromCMS(config: CMSConfig, id = false, normalizeFromJSONApi = true): CMSResults {
```
jonkoops: Doesn't seem to me `id` is always a boolean right? Should this not be an optional string or number instead?

src/app/utils/useFromCMS.ts
```diff
@@ -3,12 +3,37 @@ import { getByUrl } from '../../shared/services/api/api'
 import cmsJsonApiNormalizer from '../../shared/services/cms/cms-json-api-normalizer'
 import useNormalizedCMSResults from '../../normalizations/cms/useNormalizedCMSResults'
 
-function useFromCMS(config, id = false, normalizeFromJSONApi = true) {
-  const [results, setResults] = React.useState(null)
+export type CMSConfig = {
+  endpoint: Function
```
jonkoops: It looks to me that `endpoint` is actually a function with a signature. If I am not mistaken it should be something like:

```ts
endpoint: (id?: number) => string
```

src/app/components/CardList/CardListBlock.tsx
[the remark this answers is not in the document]
jonkoops: Ok, then let's leave it

royderks:
Updated @jonkoops

src/app/components/CardList/CardList.tsx
```diff
@@ -0,0 +1,66 @@
...
+`
+
+const StyledHeading = styled(Heading)`
+  margin: ${themeSpacing(3, 0, 6)};
+`
+
+const CardList = ({ title, list }) => {
+  const { results, fetchData, loading } = useFromCMS(list, undefined)
+
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
jonkoops: So this has been restored again, why is that?

src/app/utils/useFromCMS.ts
```diff
@@ -3,12 +3,37 @@ import { getByUrl } from '../../shared/services/api/api'
 import cmsJsonApiNormalizer from '../../shared/services/cms/cms-json-api-normalizer'
 import useNormalizedCMSResults from '../../normalizations/cms/useNormalizedCMSResults'
 
-function useFromCMS(config, id = false, normalizeFromJSONApi = true) {
-  const [results, setResults] = React.useState(null)
+export type CMSConfig = {
+  endpoint: (id?: number) => string
```
jonkoops: Looks like the id passed into this is actually a string looking at the function signature of `useFromCMS`. Is type checking disabled in your ide perhaps?

royderks: Apparently not hehe, blaim vscode

PR #2291, opened 2020-03-06
Bouwdossiers 

royderks:


Dockerfile
```diff
@@ -50,7 +50,7 @@ COPY test /app/test
 
 # Web server image
 FROM nginx:1.12.2-alpine
-ARG NODE_ENV=production
+ARG NODE_ENV=acceptance
```
royderks: Note: this should be reverted once the `bouwdossiers` API is pushed to production

Dockerfile
```diff
@@ -36,7 +36,7 @@ COPY .babelrc \
       favicon.png \
       /app/
 
-ARG NODE_ENV=production
+ARG NODE_ENV=acceptance
```
royderks: Note: this should be reverted once the `bouwdossiers` API is pushed to production

PR #2298, opened 2020-03-10
Added TileGrid component

chrisvanmook:


src/app/components/TileGrid/TileGrid.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+  Container,
+  Row,
+  themeColor,
+  themeSpacing,
+} from '@datapunt/asc-ui'
+import TileGrid from './TileGrid'
+import { TileGridItem, SizeOnBreakpoint } from './TileGridStyle'
+import { Tile, TileLabel } from '../Tile'
+
+const FullColumnTileGridItem = styled(TileGridItem)`
```
jonkoops: I think we can use a normal TileGridItem here and specify the breakpoints.

PR #2319, opened 2020-03-16
Added collection detail page

chrisvanmook:


src/app/pages/CollectionDetailPage/CollectionTileGrid.tsx
```diff
@@ -0,0 +1,131 @@
...
+  results?: CollectionTile[]
+  loading: boolean
+  title?: string
+  description?: string
+}
+
+export type CollectionTile = CMSResultItem
+
+const CollectionTileGrid: React.FC<Props> = ({ results, loading, title, description }) => {
+  type Template = {
```
jonkoops: We can pull this type out of the function

src/app/pages/CollectionDetailPage/CollectionTileGrid.tsx
```diff
@@ -0,0 +1,131 @@
...
+      { mobileL: [2, 2], tabletM: [1, 2], laptop: [1, 1] },
+    ],
+  }
+
+  const GRID_TEMPLATE: SizeOnBreakpoint = { mobileL: [1, 1], tabletM: [1, 4], laptop: [1, 5] }
+
+  return (
+    <TileGrid grid={GRID_TEMPLATE}>
+      <FullColumnTileGridItem span={{ mobileL: [2, 2] }}>
+        <Tile as="div" isLoading={loading}>
```
jonkoops: Do we need `as="div"` here?

src/app/pages/CollectionDetailPage/CollectionTileGrid.tsx
```diff
@@ -0,0 +1,131 @@
...
+          <StyledHeading forwardedAs="strong" styleAs="h1">
+            Dossier
+          </StyledHeading>
+          <Heading gutterBottom={20} as="h3" styleAs="h1">
+            {title}
+          </Heading>
+          <StyledParagraph strong>{description}</StyledParagraph>
+        </Tile>
+      </FullColumnTileGridItem>
+      {results &&
```
jonkoops: No need to check `results` here as it is always defined.

src/app/utils/useFromCMS.ts
```diff
@@ -26,10 +20,26 @@ export type CMSResultItem = {
...
+  loading: boolean
+  fetchData: Function
+  error: boolean
+}
+
+function useFromCMS<T = CMSResultItem[]>(
+  config: CMSConfig,
+  id?: string,
+  normalizeFromJSONApi = true,
+): CMSResults & {
```
jonkoops: Instead of using a type union here we can give `CMSResults` a generic type:

```ts
export type CMSResult<T> = {
  loading: boolean
  fetchData: Function
  error: boolean
  results: T[] | undefined
}

function useFromCMS<T = CMSResultItem>(): CMSResult<T> {
  // ...
}
```

Let's also make the type name signular.

src/app/utils/useFromCMS.ts
```diff
@@ -26,10 +20,26 @@ export type CMSResultItem = {
   linkProps: Object
   teaserImage?: string
   coverImage?: string
+  to?: {
```
jonkoops: I think we have a type like `Action` from Redux here that we can use, or perhaps the library we are using for links has it's on type for this one?

src/app/utils/useFromCMS.ts
```diff
@@ -46,10 +56,24 @@ function useFromCMS(config: CMSConfig, id?: string, normalizeFromJSONApi = true)
 
       let result = data
       if (normalizeFromJSONApi) {
-        result = await cmsJsonApiNormalizer(data, fields)
+        result = cmsJsonApiNormalizer(data, fields)
       }
 
-      result = useNormalizedCMSResults(result)
+      // Todo: Need to refactor this when we really know what types and fields to expect from the CMS
```
jonkoops: Yeah, this is a bit of a mess indeed. Can we create an issue to fix this so we don't lose track.

src/shared/config/cms.config.ts
```diff
@@ -96,7 +148,7 @@ const cmsConfig = {
   HOME_ORGANIZATION: {
     endpoint: () =>
       `${process.env.CMS_ROOT}jsonapi/node/list/${
-        HOMEPAGE_LINKS.ORGANIZATION.id[process.env.NODE_ENV]
+        HOMEPAGE_LINKS.ORGANIZATION.id[process.env.NODE_ENV || ENVIRONMENTS.PRODUCTION]
```
jonkoops: If we're going to do this, perhaps it makes sense to create a global module that handles this for us? Much like: https://github.com/Amsterdam/cms_search/blob/master/src/util/environment.ts

src/shared/services/cms/cms-json-api-normalizer.js
```diff
@@ -2,45 +2,49 @@ import normalize from 'json-api-normalize'
...
-        ...normalizedData,
-        ...item,
-        type: getType(item.type),
-        intro: item.field_intro,
-        short_title: item.field_short_title,
-        uuid: item.id,
-        teaser_url: teaserImageUrl,
-      }
-    })
+  // Todo: Need to refactor this when we really know what types and fields to expect from the CMS
```
jonkoops: Same for this, let's create an issue so we do not forget.

src/app/utils/useFromCMS.ts
chrisvanmook: will do 👍

src/shared/config/cms.config.ts
chrisvanmook: Let keep this out of scope for now?

src/app/pages/CollectionDetailPage/CollectionTileGrid.tsx
chrisvanmook: yeah, since it's not a link (which is the default)

src/app/components/EditorialCard/EditorialCard.jsx
```diff
@@ -57,7 +57,8 @@ const StyledLink = styled(Link)`
     }
 
     ${StyledCardContent} {
-      border-color: ${themeColor('secondary')};
+      border-color: ${({ compact }) =>
```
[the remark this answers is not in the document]
chrisvanmook: this isn't a tsx file :D

src/app/components/Tile/Tile.ts
```diff
@@ -1,36 +1,40 @@
...
   background-size: cover;
   background-repeat: no-repeat;
   padding: ${themeSpacing(12, 5)};
   min-height: 100%;
+  color: ${themeColor('tint', 'level7')};
+  background-position: center;
 
-  ${({ backgroundImageTemplate, span }) =>
-    backgroundImageTemplate && span
+  ${({ backgroundImage, span, isLoading }) =>
```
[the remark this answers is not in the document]
chrisvanmook: pfff styled component issue again... Link (which this component extends) is a TSX (react) component. If I switch to LinkStyle (which is a styled-component) it's typed... Will leave it out of scope for now, but will create an issue for this in asc

src/app/components/CardList/CardListBlock.tsx
```diff
@@ -32,30 +39,38 @@ const StyledColumn = styled(Column)`
   }
 `
 
-const CardListBlock: React.FC<{}> = () => {
-  const lists = ['List 1', 'Second list', 'Another List', 'List 4']
-
-  return (
-    <StyledRow hasMargin={false}>
+const CardListBlock: React.FC<CMSResults & { results?: CMSCollectionList[] }> = ({
```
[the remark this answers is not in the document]
chrisvanmook: I made results undefined as default in useFromCms, had type issues when using results?.field_blocks in CollectionDetailPage if results could be an empty array

src/app/components/CardList/CardList.tsx
```diff
@@ -31,47 +29,40 @@ const StyledHeading = styled(Heading)`
...
-        </div>
-      </StyledCardContent>
-    </StyledCard>
-  )
-}
+      <StyledHeading forwardedAs="h4" styleAs="h3">
+        {title}
+      </StyledHeading>
+      <div>
+        {results &&
```
jonkoops: There is no need to check the value of results here since it's always defined.

src/app/utils/useFromCMS.ts
```diff
@@ -26,10 +21,22 @@ export type CMSResultItem = {
...
-function useFromCMS(config: CMSConfig, id?: string, normalizeFromJSONApi = true): CMSResults {
-  const [results, setResults] = React.useState<CMSResultItem[]>([])
+export type CMSResults<T> = {
+  loading: boolean
+  fetchData: Function
+  error: boolean
+  results?: T
+}
+
+function useFromCMS<T = CMSResultItem[]>(
```
jonkoops: Right now when I press the 'retry' button after an error has occurred I am not seeing the loading state again. I think we need to ensure that when `fetchData` is called that the error and loading states are reset.

src/app/pages/CollectionDetailPage/CollectionDetailPage.tsx
```diff
@@ -1,7 +1,66 @@
...
+      <StyledRow>
+        {error ? (
+          <ErrorMessage onClick={() => fetchData()} />
+        ) : (
+          <>
+            <CollectionTileGrid
+              {...{
+                results: tileResults,
+                loading,
+                title: results && results.title,
```
jonkoops: Can't we write this as:

```ts
title: results?.title,
description: results?.field_intro,
```

src/app/pages/CollectionDetailPage/CollectionTileGrid.tsx
```diff
@@ -0,0 +1,129 @@
...
+    SizeOnBreakpoint,
+    SizeOnBreakpoint,
+    SizeOnBreakpoint,
+    SizeOnBreakpoint,
+  ]
+  [key: number]: SizeOnBreakpoint[]
+}
+
+const CollectionTileGrid: React.FC<Props> = ({ results, loading, title, description }) => {
+  const GRID_ITEM_TEMPLATE: Template = {
```
jonkoops: We can also pull these constants out of the function so we don't declare them again on every render.

chrisvanmook: good one 👍

PR #2320, opened 2020-03-16
Add search page for CMS collections

jonkoops:


royderks:
@jonkoops only the tests need to be updated

PR #2330, opened 2020-03-18
Bumped z-index NotificationAlert

chrisvanmook:


src/app/components/NotificationAlert/NotificationAlert.tsx
```diff
@@ -17,7 +17,7 @@ const Content = styled.div`
 `
 
 const StyledAlertMessage = styled(AlertMessage)`
-  z-index: 2;
+  z-index: 3;
```
royderks: Can't we use the z-index from the styled components repo here to avoid issues with the Header and backdrop as well?

PR #2331, opened 2020-03-18
Added link to CollectionDetailPage

chrisvanmook:


src/app/pages/CollectionDetailPage/CollectionDetailPage.tsx
```diff
@@ -56,9 +55,19 @@ const CollectionDetailPage: React.FC = () => {
               }}
             />
             <StyledCardListBlock {...{ results: listResults, loading }} />
+            {!loading && (
```
jonkoops: Should we not check `results` here instead of the loading state?

PR #2345, opened 2020-03-20
Simplify the display of content types, removed util as the types shou…

royderks:
…ld be used from Drupal directly


src/app/utils/getContentTypeLabel.ts
```diff
@@ -1,35 +0,0 @@
-import { CmsType, SpecialType } from '../../shared/config/cms.config'
-
-export default function getContentTypeLabel(type: CmsType, specialType?: SpecialType) {
```
jonkoops: Since we actually need to overwrite the label here sometimes as the type cannot be directly mapped to a human readable string let's change the implementation to return `specialType` and `type` instead of null if it cannot be matched to any cases specified here.

PR #2370, opened 2020-03-24
Refactored MapPanel & MapLegend

chrisvanmook:


src/map/components/legend/MapLegend.test.jsx
```diff
@@ -65,42 +65,4 @@ describe('MapLegend', () => {
     const component = shallow(<MapLegend {...props} />)
     expect(component).toMatchSnapshot()
   })
-
-  it('should toggle the layer', () => {
```
royderks: Do we want to deleted all the tests for this component? As the logic is quite complex..

chrisvanmook: Well tbh I think it's a bit of a waste to spend time writing tests for this now, as this component will soon be removed anyway. Moreover this solution is a POC we decided to release for now as a solution before implementing the new legend / mapPanel component. I think if we decide to add tests to this, we should also refactor the way MapLegend is working now, together with all the connected reducers, which is gonna be a hell of a job.

PR #2392, opened 2020-03-28
Using styled-components instead of asc-core

chrisvanmook:


package.json
```diff
@@ -88,6 +87,7 @@
     "redux-saga": "0.16.0",
     "remove-markdown": "0.3.0",
     "reselect": "4.0.0",
+    "styled-components": "^5.0.1",
```
jonkoops: w00t :tada:

PR #2403, opened 2020-03-30
Modified files for accessibility coming from lighthouse reports

chrisvanmook:


src/app/AppBody.jsx
```diff
@@ -46,12 +46,6 @@ const AppBody = ({
   return hasGrid ? (
     <>
       <AppContainer id="main" className="main-container">
-        <Helmet>
```
royderks: Can you check if this doesn't break the "mobile" view for pages without grid?

src/index.ejs
```diff
@@ -48,9 +48,7 @@
     <link href="<%= htmlWebpackPlugin.files.favicon %>" rel="shortcut icon">
     <% } %>
 
-    <% if (htmlWebpackPlugin.options.mobile) { %>
-    <meta content="width=device-width, initial-scale=1" name="viewport">
-    <% } %>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=5.0, minimum-scale=1.0" />
```
royderks: See above, check if this doesn't break the "mobile" page of the older components that aren't responsive

src/app/AppBody.jsx
chrisvanmook: For now I checked visiting the map page from the homepage and back, seems to work well

PR #2451, opened 2020-04-06
Combine MapLayer and MapCollection search pages

jonkoops:


src/app/pages/SearchPage/MapSearchResults/MapSearchResults.tsx
```diff
@@ -0,0 +1,119 @@
...
+  label,
+  withPagination,
+  isOverviewPage,
+}) => {
+  // Get the total count for all data types
+  const totalCount = results.reduce((acc, cur) => acc + cur.count, 0)
+  const mapLayerData = results.find(({ type }) => type === MapType.Layer)
+  const mapCollectionData = results.find(({ type }) => type === MapType.Collection)
+
+  return totalCount > 0 ? (
```
royderks, line 39: The whitespace between the layers and collections seems to be missing here:

Screenshot 2020-04-07 at 09 55 30

src/app/pages/SearchPage/MapSearchResults/MapSearchResults.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <>
+        <SearchHeading
+          label={`${label} (${totalCount.toLocaleString(DEFAULT_LOCALE)})`}
+          icon={icon}
+        />
+        <ResultsComponent results={results} />
+        <Spacer />
+        {!withPagination && !isOverviewPage && (
+          // TODO: Select proper filter.
+          <SearchLink
```
royderks, line 109: For data results we only show the link when the number of results exceeds the "first page", when on the `/data/zoek` page

src/app/pages/SearchPage/SearchPageResults.jsx
```diff
@@ -103,6 +103,11 @@ const SearchPageResults = ({
 }) => {
   const dispatch = useDispatch()
   const allResultsPageActive = currentPage === PAGES.SEARCH
+  const pageConfig = SEARCH_PAGE_CONFIG[currentPage]
+
+  if (!pageConfig) {
+    throw new Error('Unable to find page config')
```
chrisvanmook, line 109: Will this error be caught somewhere? Otherwise it'll break the whole page..

jonkoops: No, this error is not supposed to occur so we would have to find it during development.

src/app/pages/SearchPage/MapSearchResults/MapSearchResults.tsx
jonkoops: Ah, yes. I've added this check, also removed the void TODO.

src/app/pages/SearchPage/MapSearchResults/MapSearchResults.tsx
jonkoops: Fixed

src/app/pages/SearchPage/SearchPageResults.jsx
jonkoops: Changed it to not render instead

src/app/models/filter.ts
```diff
@@ -1,7 +1,7 @@
 export enum FilterType {
-  Checkbox = 'checkbox',
-  Radio = 'radio',
-  Select = 'select',
+  Checkbox = 'CHECKBOX',
```
royderks, line 2: Should this be changed back now the enum in the GraphQL schema couldn't be added?

jonkoops: Nope, not needed. I have also changed the implementation in the backend.

PR #2522, opened 2020-04-23
Various improvements to the header notifications

jonkoops:
- Use new `field_notification_title` field for the notification title
- Ensure that the element outline is visible when focussed

### DIFF
--- a/src/app/components/NotificationAlert/NotificationAlert.tsx
+++ b/src/app/components/NotificationAlert/NotificationAlert.tsx
@@ -18,6 +18,11 @@ const Content = styled.div`
 
 const StyledAlertMessage = styled(AlertMessage)`
   z-index: 3;
+
+  /* Ensure outline is visible when element is in focus */
+  &:focus {
+    z-index: 999;
+  }
 `
 
 const NotificationAlert: React.FC = () => {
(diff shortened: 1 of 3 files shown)


src/app/components/NotificationAlert/NotificationAlert.tsx
```diff
@@ -18,6 +18,11 @@ const Content = styled.div`
 
 const StyledAlertMessage = styled(AlertMessage)`
   z-index: 3;
+
+  /* Ensure outline is visible when element is in focus */
+  &:focus {
+    z-index: 999;
```
royderks, line 24: Can we use the exported z index from asc here?

src/app/components/NotificationAlert/NotificationAlert.tsx
```diff
@@ -33,14 +38,14 @@ const NotificationAlert: React.FC = () => {
...
-    const { title, body, field_notification_type, field_position } =
+    const { title, body, field_notification_type, field_position, field_notification_title } =
       results?.data[0]?.attributes || {}
 
     if (body) {
       return field_position === 'header' ? (
         <StyledAlertMessage
           dismissible
-          heading={title}
+          heading={field_notification_title}
```
royderks, line 48: Do we need a fallback to title here, same as for the InfoModal?

jonkoops: It should be left out if empty according to the specs.

src/app/components/NotificationAlert/NotificationAlert.tsx
jonkoops: Where can I find this one? I looked for `z-index` in ACS but I cannot find it.

jonkoops: Found it, but there is just one util function that doesn't accept the arguments we would need. 

This is because the header has a z-index of 21 so it's above the backdrop. If the message has a z-index below that it will fall behind the header and the focus style looks broken (as reported). If we move the message above the header it will cause conflicts since it overlaps the backdrop.

Since the overlapping backdrop and notification only happen if the user focusses a specific element using keyboard navigation and then uses the mouse to hover over the menu I think that is the lesser of two evils here.

PR #2536, opened 2020-04-28
refactored the logic for the show more link in the autosuggest, fixed…

royderks:
… bug that led to redirect to common search page on map search page


src/app/pages/SearchPage/constants.ts
```diff
@@ -0,0 +1,8 @@
+export enum SearchType {
```
chrisvanmook, line 1: A little semantical issue perhaps, but I expect this to be uppercased.

royderks: The enum or the value?

chrisvanmook: only `SEARCH_TYPE` instead of `SearchType `

chrisvanmook: oh I see it's an enum, why can't it just be a const?

royderks: It can also be a constant, but in light of migrating to TS, I created an enum. The same as has been done for MapType in example

chrisvanmook: alrighty, then ignore my remark!

PR #2559, opened 2020-04-30
Fixed netlify builds, updated graphql to v15, bumped node version

chrisvanmook:


package.json
```diff
@@ -91,11 +91,11 @@
     "urql": "^1.9.7"
   },
   "devDependencies": {
```
jonkoops, line 93: Is Dependabot not finding any updates for these?

.dependabot/config.yml
```diff
@@ -22,3 +22,5 @@ update_configs:
           dependency_name: "marked"
       - match:
           dependency_name: "body-scroll-lock"
+      - match:
+          dependency_name: "@svgr/webpack"
```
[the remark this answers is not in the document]
chrisvanmook: was suspecting this package but still seems to break..

.nvmrc
```diff
@@ -1 +1 @@
-v8.15.0
+v12.16.3
```
jonkoops, line 1: Let's use the v12 LTS instead of a fixed version here.

```suggestion
lts/erbium
```

chrisvanmook: nice! I was looking for this, but couldn't find it. Where did you got this from?

PR #2565, opened 2020-05-03
Fixed issue with alert overlay on map + fixed issue duplicated css due to css imports in js

chrisvanmook:


src/_styles.scss
```diff
@@ -1,9 +1,4 @@
 @import '~@datapunt/asc-assets/static/fonts/fonts.css';
-/* COMPONENTS */
-@import './map/map';
-
-/* Since we now use code-splitting / chunks, we need to load shared css in here to avoid warnings */
```
royderks, line 5: Is this comment resolved with this pr?

src/app/components/NotificationAlert/NotificationAlert.tsx
```diff
@@ -16,15 +16,6 @@ const Content = styled.div`
   }
 `
 
-const StyledAlertMessage = styled(AlertMessage)`
-  z-index: 3;
-
-  /* Ensure outline is visible when element is in focus */
```
royderks, line 22: Is this fix from last week still there?

src/_styles.scss
chrisvanmook: Yes, this comment is obsolete now, as we load every scss in one file now. Let's consider the .scss part legacy css, we're now moving on to styled-components

src/app/components/NotificationAlert/NotificationAlert.tsx
chrisvanmook: Reverted, but left out the z-index

PR #2569, opened 2020-05-04
Remove pano layers from the overlay when closing the panorama view

royderks:


src/panorama/sagas/panorama.js
```diff
@@ -99,7 +99,12 @@ export function* doClosePanorama() {
   const detailReference = yield select(getDetailReference)
   const pageReference = yield select(getPageReference)
   const panoramaLocation = yield select(getPanoramaLocation)
-  const overlays = yield select(getMapOverlays)
+  let overlays = yield select(getMapOverlays)
+
+  // Filter out the panorama layers, as they should be closed
+  if (overlays.length > 0) {
+    overlays = overlays.filter(({ id }) => !id.startsWith('pano'))
+  }
```
jonkoops, line 107: Even if the array is empty it should be possible to filter it, so the `if` statement could be removed.

```suggestion
overlays = overlays.filter(({ id }) => !id.startsWith('pano'))
```

Don't forget to check if the polyfill for `startsWith` is included for IE11

PR #2595, opened 2020-05-07
Added menu items

chrisvanmook:


src/app/components/HomePage/services/navigationLinks.jsx
```diff
@@ -7,9 +7,16 @@ import {
...
+  toMapSearch,
 } from '../../../../store/redux-first-router/actions'
 import { NAVIGATION_LINKS } from '../../../../shared/config/config'
+import { routing as routes } from '../../../routes'
 
+// The id's also represent the order in which they are displayed in the NavigationBlock on the homepage
+// The order of how the items are placed in the array, is the order for the Menu
 const navigationLinks = [
   {
     id: 0,
```
jonkoops, line 22: What is the point of having these ids, do they match some kind of id in a database? Otherwise why no use the index of the element in the array?

chrisvanmook: This array is used by the Menu and NavigationBlock on the homepage, either having a different order (check the comment above it).

PR #2602, opened 2020-05-08
Use the updated features  on the matomo-tracker package to improve an…

royderks:
…alytics tracking  for the project


src/app/pages/ArticleDetailPage/ArticleDetailPage.jsx
```diff
@@ -111,6 +111,12 @@ const StyledAccordion = styled(Accordion)`
   margin-top: ${themeSpacing(2)};
 `
 
+const StyledLink = styled(Link)`
+  & > * {
```
royderks, line 115: Should we update in asc-ui? We want clicks on the link to be registered on a click on the link, and not by a link on the inner elements and have the click event bubble. The link should be the only clickable element, unless specifically overwritten by the user of the library

royderks: @chrisvanmook @jonkoops

jonkoops: Seems a bit like a hack IMHO, why isn't the analytics handling this properly?

src/app/AppBody.jsx
```diff
@@ -42,6 +43,10 @@ const AppBody = ({
   embedPreviewMode,
 }) => {
   const query = useSelector(getQuery)
+
+  const { enableLinkTracking } = useMatomo()
```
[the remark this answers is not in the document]
royderks: The useEffect is already handled by the package, where the event handler is added.

src/app/pages/ArticleDetailPage/ArticleDetailPage.jsx
royderks: I believe it's due to how the application is being rendered, that the event can't be caught. The target element is the inner span while the currentTarget is the entire DOM (without the rendered html elements rendered by React)

src/app/AppBody.jsx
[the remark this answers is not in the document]
jonkoops: Right, but we still do not want to call `enableLinkTracking` on every render right?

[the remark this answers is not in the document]
royderks: It doesn't really matter, as the `AppBody` component is rerendered on every route change, it will be called either way

PR #2603, opened 2020-05-09
Refactored code to be used in a server side environment

chrisvanmook:


src/app/components/DataSelection/DataSelection.jsx
```diff
@@ -15,9 +15,15 @@ import DataSelectionTable from './DataSelectionTable/DataSelectionTable'
 import DataSelectionList from './DataSelectionList/DataSelectionList'
 import ShareBar from '../ShareBar/ShareBar'
 import Notification from '../../../shared/components/notification/Notification'
-import '../../angularModules'
 import { DEFAULT_LOCALE } from '../../../shared/config/locale.config'
 
+let angularInstance = null
+
+if (typeof window !== 'undefined') {
```
jonkoops, line 22: Instead of using instanceof you can also compare the value to undefined directly, eg.

```ts
if (window !== undefined) {
```

src/store/middleware/matomo/matomoMiddleware.js
```diff
@@ -6,6 +6,10 @@ import matomoInstance from '../../../app/matomo'
 // Execute Matomo actions
 const matomoMiddleware = ({ getState }) => (next) => (action) => {
   const nextAction = action
+  if (typeof window === 'undefined') {
```
jonkoops, line 9: Should this be wrapped in a conditional? I don't see the window being used here.

src/store/middleware/documentHead.js
```diff
@@ -21,7 +21,9 @@ const documentHead = () => (next) => (action) => {
     const getTitle = get(titleResolver, 'getTitle', getDefaultDocumentTitle(page))
     const pageTitle = getTitle(action, getDefaultDocumentTitle(page)())
 
-    document.title = `${pageTitle} - ${TITLE_SUFFIX}`
+    if (typeof window !== 'undefined') {
```
jonkoops, line 24: Instead of checking the window here would it not make more sense to check for the existance of `document` instead?

src/app/components/DataSelection/DataSelection.jsx
chrisvanmook: That will still throw a reference error, as `window` isn't declared on the server.

src/store/middleware/matomo/matomoMiddleware.js
chrisvanmook: on line 26 and 27 it's used :)

src/store/middleware/documentHead.js
chrisvanmook: sure, however I can't figure out a situation where `document` isn't defined but `window` is

src/app/components/DataSelection/DataSelection.jsx
jonkoops: Really? You'd expect that would only happen if we accessed a member of window itself

chrisvanmook: image

jonkoops: Lolwhat. Well good to know :man_shrugging:

PR #2610, opened 2020-05-11
Dont close the filterbox when changing the base path of the search page

royderks:


src/store/parameters.js
```diff
@@ -13,7 +13,6 @@ export default {
   PANORAMA_TAGS: 'tags',
   PITCH: 'pitch',
   FILTERS: 'filters',
-  SEARCH_FILTERS: 'zoek-filters',
```
royderks, line 16: Deleted this as it was unused

PR #2637, opened 2020-05-18
Feature/open map panel

royderks:


src/map/components/legend/MapLegend.jsx
```diff
@@ -196,6 +196,21 @@ const MapLegend = ({
...
   }
 
+  useEffect(() => {
+    // Open the panel when some of the layers are visible
+    if (
+      !isOpen &&
+      mapLayers.some(
+        ({ isVisible, legendItems }) =>
+          isVisible === true ||
+          (isVisible === true &&
```
jonkoops, line 206: Do you need this second explicit check?

royderks: Nope, deleted

PR #2641, opened 2020-05-18
Improve build system configuration

jonkoops:
Changes:
- WebPack config files are now written in TypeScript
- Polyfills are added automatically through babel-env
- Dependencies written in ES Modules now supported
- Seperate build for legacy platforms such as Internet Explorer (differential builds)
- Modern build shipping with ES2015 code for modern platforms


src/app/AppBody.jsx
```diff
@@ -12,21 +12,41 @@ import LoadingIndicator from '../shared/components/loading-indicator/LoadingIndi
...
-const HomePage = React.lazy(() => import('./pages/HomePage'))
-const ActualityContainer = React.lazy(() => import('./containers/ActualityContainer'))
+const HomePage = React.lazy(() => import(/* webpackChunkName: "HomePage" */ './pages/HomePage'))
+const ActualityContainer = React.lazy(() =>
+  import(/* webpackChunkName: "ActualityContainer" */ './containers/ActualityContainer'),
+)
 const DatasetDetailContainer = React.lazy(() =>
-  import('./containers/DatasetDetailContainer/DatasetDetailContainer'),
+  import(
+    /* webpackChunkName: "DatasetDetailContainer" */ './containers/DatasetDetailContainer/DatasetDetailContainer'
```
chrisvanmook, line 21: Are we chunking all pages?

webpack.common.ts
```diff
@@ -0,0 +1,275 @@
...
+          },
+        },
+        {
+          test: /\.svg$/,
+          use: [
+            {
+              loader: '@svgr/webpack',
+              options: {
+                svgo: options.mode === 'production',
+                svgoConfig: {
```
chrisvanmook, line 141: same config as above, any way we can put this in a variable?

webpack.prod.ts
```diff
@@ -1,32 +1,39 @@
...
-      'leaflet|leaflet-draw|leaflet-rotatedmarker|leaflet.markercluster|leaflet.nontiledlayer|proj4|proj4leaflet',
-    DATAPUNT: '@datapunt',
-    STYLED: 'styled-components|polished|style-loader|css-loader|sass-loader|postcss-loader',
-    PANORAMA: 'marzipano',
-    POLYFILL: '@babel/polyfill|objectFitPolyfill|core-js',
-    ANGULAR: 'angular|angular-aria|angular-sanitize|react-angular',
-    REACT:
-      'react|react-dom|redux-first-router|redux-first-router-link|redux-first-router-restore-scroll|reselect|redux|@?redux-saga|react-redux|react-helmet|prop-types',
-  }
+const CHUNKS = {
```
chrisvanmook, line 6: Isn't this gonna clash with the lazy imports?

src/app/AppBody.jsx
jonkoops: Most of them seem to be split into seperate lazy-loaded chunks. There are some optimizations that we can still do looking at the bundle, but I consider that out of scope for this PR. I have however created an issue on our backlog so we can take a look at this later.

webpack.prod.ts
jonkoops: It shouldn't. The chunks should load in whenever a module that requires them is loaded.

webpack.common.ts
jonkoops: We could, but I don't see the advantage of it as this config is only used once.

chrisvanmook: It's used here as well, right? https://github.com/Amsterdam/atlas/pull/2641/files#diff-ff4fc69cd4b6201389c9738391bd9e0eR220

jonkoops: Oh, yeah nice catch. I merged the two definitions.

PR #2648, opened 2020-05-19
Replace the alert messages with the one from asc-ui, to better match …

royderks:
…accessibility requirements


src/app/models/notification.ts
```diff
@@ -0,0 +1,8 @@
+export enum NotificationLevel {
```
jonkoops, line 1: Is there a need for this enum? The styled component for `Alert` should already only accept these types since it's using a string union.

royderks: It could also be a constant if you'd prefer that

jonkoops: I'd prefer the string literals, but it's more of a pet peeve honestly.

PR #2678, opened 2020-05-26
Beta map

royderks:


src/app/pages/MapPage/MapContainer.tsx
```diff
@@ -0,0 +1,307 @@
...
+import MapContext, { MapContextProps, MapStateProps, initialState } from './MapContext'
+import MAP_CONFIG from '../../../map/services/map.config'
+import { createUrlWithToken } from '../../../shared/services/api/api'
+
+function setQueryStringParameter(name: string, value: string) {
+  const params = new URLSearchParams(window.location.search)
+
+  params.set(name, value)
+
+  window.history.pushState({}, '', decodeURIComponent(`${window.location.pathname}?${params}`))
```
chrisvanmook, line 20: Where is the logic where we want to replaceState instead of pushing it? eg. when drawing and edit the drawing, it should replace the state

src/app/pages/MapPage/Components/DrawTool.tsx
```diff
@@ -0,0 +1,146 @@
...
+      if (mapInstance) {
+        // @ts-ignore
+        mapInstance.off(L.Draw.Event.EDITVERTEX, editVertex)
+      }
+    }
+  }, [mapInstance])
+
+  return (
+    <DrawToolComponent
+      onDrawEnd={async (layer: ExtendedLayer) => {
```
chrisvanmook, line 129: We should dispatch an action from here to sync with the url

src/app/pages/MapPage/MapContainer.tsx
```diff
@@ -0,0 +1,307 @@
...
+        {},
+      ),
+    })
+  }
+
+  async function getPanelLayers() {
+    const panelLayers = await fetchPanelLayers()
+
+    dispatch({
+      type: 'getPanelLayers',
```
chrisvanmook, line 150: could we use the same convention we have in de "old" reducers? In constants, capitalised and prefixed with reducer name (if needed, not sure if it's already scoped when using react reducers)

royderks: Not needed, as it's scoped to the provider the components are wrapped in

src/app/pages/MapPage/Components/DrawTool.tsx
royderks: Yes I've updated this

src/app/pages/MapPage/MapContainer.tsx
royderks: When should we replace state? We want to be able to go back right?

chrisvanmook: well if you toggle map layers for example, every time you toggle the URL should be updated. If you toggle a lot of layers each time, you create a huge history.

src/app/pages/MapPage/MapContainer.tsx
```diff
@@ -0,0 +1,341 @@
...
+      dispatch({ type: 'setGeometry', payload })
+    }
+  }
+
+  function toggleMapPanel() {
+    setMapPanelVisible(!state.isMapPanelVisible)
+  }
+
+  async function getBaseLayers() {
+    const baseLayers = (await fetchBaseLayers()) as Array<{ category: string }> // Add auto type generation
```
jonkoops, line 192: If we have some idea of what this type looks like I would put it in shared code so that we only have to swap that single type instead. Otherwise it's going to be hard to refactor this code, which is what TypeScript is great for.

src/app/pages/MapPage/MapContainer.tsx
```diff
@@ -0,0 +1,341 @@
...
+      return state
+  }
+}
+
+const MapContextProvider: React.FC<MapContextProps & { children: React.ReactNode }> = ({
+  children,
+}) => {
+  const [state, dispatch] = React.useReducer(reducer, initialState)
+
+  function setActiveBaseLayer(payload: string) {
```
jonkoops, line 150: Instead of defining these here in the component would it not make more sense to put these methods outside and dispatch the result of them? E.g.

```ts
const setActiveBaseLayer = (payload: string) => ({ type: 'setActiveBaseLayer', payload })
	

dispatch(setActiveBaseLayer('test'))
```

src/app/pages/MapPage/Components/GeoJSON.tsx
```diff
@@ -0,0 +1,24 @@
...
+  geometry: Geometry
+}
+
+const GeoJSON: React.FC<Props> = ({ geometry }) => {
+  console.log('hiiiiii', geometry)
+
+  return (
+    <RDGeoJSON
+      geometry={geometry}
+      //   options={{ style: defaultStyle }}
```
jonkoops, line 19: Leftover comment

src/app/pages/MapPage/Components/DrawTool.tsx
```diff
@@ -0,0 +1,146 @@
+import L, { LatLng, LatLngTuple, Polygon, Polyline } from 'leaflet'
+import React, { useEffect, useState } from 'react'
+import { components } from '@datapunt/amsterdam-react-maps'
```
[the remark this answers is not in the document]
chrisvanmook: This is something we should fix in amsterdam-react-maps

src/app/pages/MapPage/Components/DrawTool.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    if (layer instanceof Polygon) {
+      toolTipText = `${L.GeometryUtil.readableArea(L.GeometryUtil.geodesicArea(latLngs), true, {
+        m: 1,
+      })}, ${toolTipText}`
+    }
+    if (toolTipText) {
+      layer.bindTooltip(toolTipText, { direction: 'bottom' }).openTooltip()
+    }
+  }
```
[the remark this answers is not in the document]
chrisvanmook: I will pick this up

src/map/containers/panel/MapPanel.jsx
```diff
@@ -59,7 +58,7 @@ const MapPanel = ({
     <div className="scroll-wrapper">
       <MapPanelHandle
         {...{
-          isMapPanelHandleVisible,
+          isMapPanelHandleVisible: true, // TODO: find out if we actually need these (and from where its used)
```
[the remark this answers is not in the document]
royderks: Later, as this component will be replaced by the new maplegend

PR #2726, opened 2020-06-02
Refactored the backdrop to be filling the entire page for the menu an…

royderks:
…d the autosuggest

### DIFF
--- a/src/app/components/Header/HeaderMenu.jsx
+++ b/src/app/components/Header/HeaderMenu.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { MenuInline, MenuToggle, MenuFlyOut, MenuItem, MenuButton } from '@datapunt/asc-ui'
+import { MenuFlyOut, MenuItem, MenuButton } from '@datapunt/asc-ui'
 import { ChevronRight } from '@datapunt/asc-assets'
 import PropTypes from 'prop-types'
 import { useMatomo } from '@datapunt/matomo-tracker-react'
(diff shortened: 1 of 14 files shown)


src/app/components/Header/HeaderMenu.jsx
```diff
@@ -10,17 +10,11 @@ import navigationLinks from '../HomePage/services/navigationLinks'
 
 import { HEADER_LINKS } from '../../../shared/config/config'
 
-const components = {
-  default: MenuInline,
-  mobile: MenuToggle,
-}
-
-const HeaderMenu = ({ type, login, logout, user, showFeedbackForm, ...props }) => {
+const HeaderMenu = ({ Component, login, logout, user, showFeedbackForm, ...props }) => {
```
jonkoops, line 13: It looks a bit weird that this is a prop that is capitalized, the name is also a bit non-descriptive. Perhaps we can rename the prop in this function?

src/app/components/Header/HeaderMenu.jsx
```diff
@@ -137,15 +131,15 @@ const HeaderMenu = ({ type, login, logout, user, showFeedbackForm, ...props }) =
           </MenuItem>
         </MenuFlyOut>
       )}
-    </Menu>
+    </Component>
   )
 }
 
 HeaderMenu.propTypes = {
```
jonkoops, line 138: Seems an easy file for a a TypeScript conversion :)

PR #2727, opened 2020-06-02
Open the maplegend when some of the maplayers are selected on the fir…

royderks:
…st render


src/map/components/legend/MapLegend.jsx
```diff
@@ -170,6 +165,18 @@ const MapLegend = ({
     }
   }
 
+  // Open the MapLegend and scroll it into the current view
+  const handleSetOpen = (open) => {
```
jonkoops, line 169: We can probably wrap this in a `useCallback`

royderks: Yes sure, it was already there but can be optimized

PR #2734, opened 2020-06-03
Added BaseLayerToggle component to beta map

chrisvanmook:


src/app/pages/MapPage/MapPage.tsx
```diff
@@ -133,6 +139,14 @@ const MapPage: React.FC = () => {
         ))}
         <StyledViewerContainer
           bottomRight={isLoading ? <Spinner /> : null}
+          bottomLeft={
+            <BottomLeftHolder>
+              <MapPanelContainerWrapper>
+                <MapPanelContainer />
+              </MapPanelContainerWrapper>
+              <BaseLayerToggle />
```
royderks, line 147: How can we get the selection from the baselayer toggle? There already is logic to sync that selection with the url

PR #2741, opened 2020-06-04
Feature/iiif files

royderks:


src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
```diff
@@ -0,0 +1,151 @@
...
+  padding: ${themeSpacing(1, 5)};
+  white-space: normal;
+  font-weight: 500;
+
+  &:first-child {
+    width: 33%;
+  }
+`
+
+const ConstructionFileDetail: React.FC<{ results: ConstructionFileDetailProps }> = ({
```
jonkoops, line 76: I would expect `ConstructionFileDetailProps` to be passed in here directly.

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
```diff
@@ -0,0 +1,151 @@
...
+  subdossier_titel: string
+  access: 'RESTRICTED' | 'PUBLIC'
+}
+
+type ConstructionFileAddress = {
+  id: string
+  label: string
+}
+
+type ConstructionFileDetailProps = {
```
jonkoops, line 27: Looking at the code this seems to be a 'result' in the component. Perhaps it would be better to call this one  `ConstructionFileDetailResult` and create a new type that is called `ConstructionFileDetailProps` that contains it?

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
```diff
@@ -0,0 +1,151 @@
...
+          <TableCell>Type</TableCell>
+          <TableCell>{fileType}</TableCell>
+        </TableRow>
+        <TableRow>
+          <TableCell>Dossiernummer</TableCell>
+          <TableCell>{fileNumber}</TableCell>
+        </TableRow>
+      </Table>
+
+      {documents &&
```
jonkoops, line 116: According to the type definition the `documents` field should always be defined so we can skip these checks and map the array directly.

src/app/components/Gallery/Gallery.tsx
```diff
@@ -0,0 +1,161 @@
...
+  margin-right: ${themeSpacing(4)};
+  margin-bottom: ${themeSpacing(4)};
+`
+
+const StyledLink = styled(Link)`
+  width: 100%;
+  height: 100%;
+  position: relative;
+
+  ${({ disabled }) =>
```
jonkoops, line 43: Is disabled already typed? Or should we add a type definition here?

src/app/components/Gallery/Gallery.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  const hasMore = allFiles.length > MAX_LENGTH
+  const restricted = access === 'RESTRICTED'
+
+  return (
+    <GalleryContainer key={title}>
+      <StyledHeading color="secondary" forwardedAs="h3">
+        {title} {hasMore && `(${allFiles.length})`}
+      </StyledHeading>
+      {files && files.length ? (
```
jonkoops, line 90: According to the type definition `files` is always defined. So this check can either be removed or the type is incorrect.

src/app/components/Gallery/Gallery.tsx
```diff
@@ -0,0 +1,161 @@
...
+            {files.map(({ filename: fileName, url: fileUrl }) => {
+              const disabled = (!hasRights && !hasExtendedRights) || (restricted && !hasRights)
+
+              return (
+                <StyledColumn
+                  key={fileName}
+                  span={{ small: 1, medium: 2, big: 2, large: 2, xLarge: 2 }}
+                >
+                  {/*
+                  // @ts-ignore */}
```
jonkoops, line 117: What is this `ts-ignore` needed for?

royderks: The `StyledLink` doesn't accept `disabled` as a prop

royderks:
Updated @jonkoops!

src/app/components/Gallery/Gallery.tsx
jonkoops: Ah, ok. Yeah, and RouterLink does, makes sense.

PR #2775, opened 2020-06-10
Add type information to User duck and saga

jonkoops:


src/shared/ducks/user/user.ts
```diff
@@ -0,0 +1,168 @@
...
+  name: string,
+  scopes: string[],
+  reload: boolean,
+): AuthenticateUserSuccessAction => {
+  const meta = !reload ? { tracking: scopes } : null
+
+  return {
+    type: AUTHENTICATE_USER_SUCCESS,
+    payload: { accessToken, name, scopes },
+    ...meta,
```
jonkoops, line 158: I think this might be a bug that already existed, as other actions have `meta` as a seperate field. Should I change this to match the others, or perhaps just remove it? What do you think @chrisvanmook @royderks ?

PR #2795, opened 2020-06-15
Add type checks to the compiler and fix typing issues

jonkoops:


webpack.common.ts
```diff
@@ -271,6 +272,14 @@ export function createConfig(additionalOptions: CreateConfigOptions): Configurat
             : false,
       }),
       ...(!options.singleBuild ? [new HtmlWebpackMultiBuildPlugin()] : []),
+      new ForkTsCheckerWebpackPlugin({
```
royderks, line 275: Do we also need this in production? Or is it sufficient to include it in develop only?

jonkoops: I think it makes sense to include this in production as well as we want to run these builds on the CI.

jonkoops:
Atlas has a lot of type errors, so I'll update this PR with some fixes, `ts-ignore` and `any` types to make the build pass. Will do this when I have some time as this is not our main concern right now.

jonkoops:
I'll be opening seperate PRs to update the type compatibility and rebasing this PR with develop until the build succeeds.

src/app/pages/MapPage/MapPage.tsx
```diff
@@ -137,15 +130,7 @@ const MapPage: React.FC = () => {
             />
           ))}
           {nonTmsLayers.map(({ url, overlayOptions: options, id }) => (
-            <NonTiledLayer
-              key={id}
-              url={url}
-              options={options}
-              events={{
```
chrisvanmook, line 144: why is this removed?

src/app/pages/MapPage/MapPage.tsx
```diff
@@ -172,15 +157,17 @@ const MapPage: React.FC = () => {
                 )}
                 {!showDrawTool && !detailUrl && location && (
                   <PointSearchPanel
-                    {...{
```
chrisvanmook, line 175: why is this replaced? something to do with typechecking?

chrisvanmook:
I presume that you tested the app thoroughly right? I can't really estimate what the impact is of all these changes...

src/app/components/CardList/CardList.tsx
```diff
@@ -54,18 +53,15 @@ const CardList: React.FC<CardListProps> = ({ title, loading, results }) => (
             coverImage,
           }) => (
             <EditorialCard
-              {...{
-                ...linkProps, // !important: linkProps also contains `title`
```
royderks, line 58: Did you check if the title is still displayed correct?

jonkoops: Yes, the `title` prop is overwritten further down.

src/app/pages/MapPage/MapPage.tsx
jonkoops: Yeah, if you spread an object you cannot easily see which individual properties are invalid.

src/app/pages/MapPage/MapPage.tsx
jonkoops: I checked `NonTiledLayer` and it looks like it does not have an `events` prop (see: https://github.com/Amsterdam/amsterdam-react-maps/blob/master/packages/arm-nontiled/src/components/NonTiledLayer.tsx#L6-L10)

jonkoops:
> I presume that you tested the app thoroughly right? I can't really estimate what the impact is of all these changes...

Yes, I've tested almost all of the code except small changes and it seems to work all right.

PR #2826, opened 2020-06-19
Add a definition list component that can be used on detail pages

royderks:


src/app/components/DefinitionList/DefinitionList.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { themeSpacing } from '@datapunt/asc-ui'
+import styled from 'styled-components'
+
+const DefinitionListWrapper = styled.ul`
```
chrisvanmook, line 5: I'd expect a DL element when I would use a DefinitionList. Could we rename or actually use DL ?

jonkoops: Semantically this would be the better option

src/app/components/DefinitionList/DefinitionList.tsx
```diff
@@ -0,0 +1,14 @@
...
+  width: 100%;
+  margin-bottom: ${themeSpacing(2)};
+`
+
+const DefinitionList: React.FC<{}> = ({ children }) => (
```
jonkoops, line 10: If the component takes no props you can also drop the generic type to `React.FC`.

```suggestion
const DefinitionList: React.FC = ({ children }) => (
```

src/app/components/DefinitionList/DefinitionListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+  width: 30%;
+`
+
+const DefinitionListValue = styled.div`
+  padding: ${themeSpacing(1, 5)};
+  white-space: normal;
+  width: 70%;
+`
+
+const DefinitionListItem: React.FC<{ label: string; value: string | number }> = ({
```
jonkoops, line 30: Instead of in-lining these props I would recommend to pull them into a separate exported interface so that it could be re-used or extended by other code.

```ts
export interface DefinitionListItemProps {
   label: string
   value: string | number
}
```

royderks:
Updated @jonkoops @chrisvanmook

PR #2834, opened 2020-06-22
Add results to point search panel

jonkoops:


src/app/pages/MapPage/Components/PointSearchResults.tsx
```diff
@@ -1,49 +1,152 @@
...
-          quam quibusdam sapiente similique sit suscipit ut vel vero! Accusamus ad consequatur
-          dolore esse, facere fugiat illum maxime mollitia nihil optio, quasi quisquam reprehenderit
-          saepe sunt totam unde vel veniam veritatis vero voluptates.
-        </Paragraph>
-      )}
-    </MapPanelContent>
+      <CoordinatesText>
+        <strong>Locatie:</strong> {location.lat}, {location.lng}
+      </CoordinatesText>
+      <PanoramaPreview location={location} radius={180} aspect={2.5} />
```
royderks, line 99: Should we define the radius and aspect somewhere? Or have defaults for the panoramapreview component?

src/app/pages/MapPage/Components/PointSearchResults.tsx
```diff
@@ -1,49 +1,152 @@
...
+      ))}
+    </ShowMore>
+  )
+}
+
+const mapStateToProps = (state: RootState) => ({
+  user: getUser(state),
+})
+
+export default connect(mapStateToProps)(PointSearchResults)
```
royderks, line 152: Do we still want to use the HOC for redux here, or already decouple a bit more and use the `useSelector` hook from react-redux? Also, there is a getState function in the codebase to get the redux code with a HOC

src/app/pages/MapPage/Components/PointSearchResults.tsx
jonkoops: I think it's fine to keep these in-line here, the API itself already provides defaults so I'd rather not override these.

src/app/pages/MapPage/Components/PointSearchResults.tsx
jonkoops: I wasn't aware of this hook, but it looks much better. I've changed the code to use it instead.

PR #2898, opened 2020-07-03
Allow users to draw and save multiple polygons, delete the polygon st…

royderks:
…ate when closing the draw tool

### DIFF
--- a/src/app/pages/MapPage/Components/DrawTool.tsx
+++ b/src/app/pages/MapPage/Components/DrawTool.tsx
@@ -5,15 +5,16 @@ import {
   PolygonType,
   PolylineType,
 } from '@datapunt/arm-draw'
-import { ascDefaultTheme, themeColor } from '@datapunt/asc-ui'
+import { themeColor, ascDefaultTheme } from '@datapunt/asc-ui'
+import L, { Polygon, LatLng, LatLngTuple, LatLngLiteral } from 'leaflet'
 import { useMapInstance } from '@datapunt/react-maps'
-import L, { LatLng, LatLngLiteral, LatLngTuple, Polygon } from 'leaflet'
+
 import React, { useContext, useEffect, useMemo } from 'react'
 import PARAMETERS from '../../../../store/parameters'
 import { decodeBounds } from '../../../../store/queryParameters'
 import getParam from '../../../utils/getParam'
 import DataSelectionContext from '../DataSelectionContext'
-import MapContext from '../MapContext'
+import MapContext, { DrawingGeometry } from '../MapContext'
 import { Overlay, SnapPoint } from '../types'
 
 type MarkerGroup = {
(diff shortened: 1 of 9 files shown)


src/app/pages/MapPage/Components/DrawTool.tsx
```diff
@@ -162,12 +160,15 @@ const DrawTool: React.FC<Props> = ({ onToggle, isOpen, setCurrentOverlay }) => {
     return polybounds.length > 2 ? createPolygon(polybounds) : createPolyline(polybounds)
   }
 
-  const initalDrawnItem = useMemo(() => {
+  const initalDrawnItems = useMemo(() => {
```
chrisvanmook, line 163: typo

src/app/pages/MapPage/Components/DrawTool.tsx
```diff
@@ -162,12 +160,15 @@ const DrawTool: React.FC<Props> = ({ onToggle, isOpen, setCurrentOverlay }) => {
...
     // Get the drawing geometry from the url here, before the context is updated to prevent newly drawm geometries to be pushed to the DrawToolComponent
-    const initialDrawingGeometry = getParam(PARAMETERS.DRAWING_GEOMETRY)
-
-    if (initialDrawingGeometry) return setInitialDrawing(decodeBounds(initialDrawingGeometry))
+    const initialDrawingGeometries = getParam(PARAMETERS.DRAWING_GEOMETRY)
 
+    if (initialDrawingGeometries) {
+      return decodeBounds(initialDrawingGeometries).map((initialDrawingGeometry: DrawingGeometry) =>
+        setInitialDrawing(initialDrawingGeometry),
```
chrisvanmook, line 169: shouldn't we move this to a useEffect hook?

royderks: Why? These geometries should only be created on a render of the component, so the useMemo can be used instead. Otherwise a local state should be created to hold these geometries that can be updated, but won't be updated as there is no need for it

src/app/pages/MapPage/MapContainer.tsx
```diff
@@ -277,22 +291,28 @@ const MapContextProvider: React.FC<MapContextProps> = ({ children }) => {
...
-    if (drawingGeometry) setDrawingGeometry(decodeBounds(drawingGeometry))
-  }, [baseLayer, activeMapLayers, location, detailUrl, drawingGeometry])
+  }, [baseLayer, activeMapLayers, location, detailUrl])
 
   // This can be refactored if the mapLayers are added in batches
   React.useEffect(() => {
     if (state.activeMapLayers.length) setParam('lagen', encodeLayers(state.activeMapLayers))
   }, [state.activeMapLayers])
 
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
royderks: Yes that would be a good idea, but hard to do as the url state depends on the lifecycles in the different components that use that state

PR #2900, opened 2020-07-06
Add the value of the dropdown in the searchbarfilter to the session s…

royderks:
…torage

### DIFF
--- a/src/app/components/SearchBarFilter/SearchBarFilter.tsx
+++ b/src/app/components/SearchBarFilter/SearchBarFilter.tsx
@@ -1,4 +1,4 @@
-import React, { Dispatch, SetStateAction } from 'react'
+import React, { Dispatch, SetStateAction, useEffect, useState } from 'react'
 import { Select, Label } from '@datapunt/asc-ui'
 import styled from 'styled-components'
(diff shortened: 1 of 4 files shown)


src/app/components/SearchBarFilter/SearchBarFilter.tsx
```diff
@@ -25,30 +25,46 @@ type SearchBarFilterProps = {
...
+
 const SearchBarFilter: React.FC<SearchBarFilterProps> = ({ searchCategory, setSearchCategory }) => {
-  const AVAILABLE_FILTERS = Object.keys(SEARCH_PAGE_CONFIG).map((key) => {
-    const { type, label } = SEARCH_PAGE_CONFIG[key]
-    return {
-      type,
-      label,
-    }
-  })
+  const AVAILABLE_FILTERS: Array<SearchBarFilterOptions> = Object.keys(SEARCH_PAGE_CONFIG).map(
```
jonkoops, line 34: Since this is constructed as a constant I think we can move it outside of the component. Also since we're accessing the values directly can't we use `Object.values` instead?

royderks: Yes, nicely spotted

PR #2901, opened 2020-07-06
Added IE11 deprecation Alert

chrisvanmook:


src/app/AppBody.jsx
```diff
@@ -57,6 +58,15 @@ const AppContainer = styled.div`
...
 `
 
+const StyledAlert = styled(Alert)`
+  /* Ensure outline is visible when element is in focus */
+  &:focus {
+    z-index: 999;
+  }
+`
+
+const IS_IE11 = !!window.document.documentMode
```
jonkoops, line 68: We already have this check in another part of the application, perhaps we can make it re-usable?

PR #2910, opened 2020-07-06
Refactor LoginLink component to be type safe

jonkoops:


src/app/components/Links/LoginLink/LoginLink.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const LoginLink: React.FC<LoginLinkProps> = ({ variant = 'with-chevron', children }) => {
+  const dispatch = useDispatch()
+  const handleLogin = useCallback((event: MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault()
+    dispatch(authenticateRequest('inloggen'))
+    login()
+  }, [])
+
+  // TODO: Since this is not a navigational element a Button should be used here.
```
royderks, line 27: Can't you use the forwardedAs prop?

jonkoops: We can but since that makes the element type polymorphic it cannot be written in a type-safe manner. I think we should create a new `LinkButton` component that acts exactly like the `Link` element in terms of appearance.

I've created an issue so we can implement this in the future: https://github.com/Amsterdam/amsterdam-styled-components/issues/794

PR #2919, opened 2020-07-07
Allow users to draw and save multiple polygons

jonkoops:


src/app/pages/MapPage/Components/DrawTool.tsx
```diff
@@ -126,6 +122,31 @@ const DrawTool: React.FC<Props> = ({ onToggle, isOpen, setCurrentOverlay }) => {
     await getData(layer, distanceText)
   }
 
+  const onDeleteDrawing = (layersInEditMode: Array<ExtendedLayer>) => {
```
chrisvanmook, line 125: Perhaps when we will refactor the URL state logic we could introduce an observer pattern or sth like that? Because this method is doing a lot... @royderks @jonkoops

royderks: Sounds good, I've added it to the ticket as reference

PR #2922, opened 2020-07-07
Allow clicking of items on the map

jonkoops:


src/map/services/nearest-detail/nearest-detail.js
```diff
@@ -44,5 +44,5 @@ export default async function fetchNearestDetail(location, layers, zoom) {
       )
     ).reduce(/* istanbul ignore next */ (a, b) => [...a, ...b]),
   )
-  return results[0] ? results[0] : ''
+  return results[0] ? results[0] : null
```
royderks, line 47: `results[0] || null` ?

PR #2934, opened 2020-07-09
Unify enviroment specific NGINX configs

jonkoops:


jonkoops:
We should remove the `DEPLOY_ENV` from the build pipelines

PR #2940, opened 2020-07-13
Various bugfixes for dynamic environment

jonkoops:
- Pass VERSION and GIT_COMMIT environment variables into the build
- Place startup script into /app/ directory in Docker build
- Update application information message with GIT_COMMIT
- Pass GIT_COMMIT into Sentry calls

### DIFF
--- a/src/app/index.jsx
+++ b/src/app/index.jsx
@@ -1,11 +1,11 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
 import { Provider } from 'react-redux'
-import App from './App'
+import environment from '../environment'
 import configureStore from '../store/store'
+import App from './App'
 import resolveRedirects from './redirects'
 import './sentry'
-import environment from '../environment'
 
 if ('serviceWorker' in navigator) {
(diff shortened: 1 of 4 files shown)


src/app/index.jsx
```diff
@@ -26,8 +26,12 @@ resolveRedirects().then((hasToRedirect) => {
 })
 
 function renderApp(store) {
-  // eslint-disable-next-line no-undef,no-console
-  console.log(`CityData: version: ${VERSION}, build: ${environment.DEPLOY_ENV}`)
+  // eslint-disable-next-line no-console
+  console.log(
+    `Dataportaal: version: ${process.env.VERSION}, deploy env: ${environment.DEPLOY_ENV}${
```
royderks, line 31: Can you describe somewhere when to use `environment` and when to use `process.env.*`?

jonkoops: `proccess.env.*` is for environment variables which are constructed at build time and are provided in the Webpack config. `environment` has the environment variables which are defined at boot time by the Docker container.

PR #2945, opened 2020-07-13
Make construction files and related code type-safe

jonkoops:


src/app/components/ImageViewer/ImageViewer.tsx
```diff
@@ -159,7 +159,6 @@ const ImageViewer: React.FC<ImageViewerProps> = ({
             <ContextMenu
               handleDownload={handleDownload}
               downloadLoading={downloadLoading}
-              fileName={fileName}
```
royderks, line 162: Why is this prop removed? The users want to download the file with a the original filename

src/app/containers/ActualityContainer.jsx
```diff
@@ -1,7 +1,13 @@
...
-import LoadingIndicator from '../../shared/components/loading-indicator/LoadingIndicator'
-import { fetchWithoutToken } from '../../shared/services/api/api'
+import styled from 'styled-components'
 import environment from '../../environment'
+import { fetchWithoutToken } from '../../shared/services/api/api'
+import LoadingSpinner from '../components/LoadingSpinner/LoadingSpinner'
+
+const StyledLoadingSpinner = styled(LoadingSpinner)`
+  position: absolute;
+  top: 200px;
```
royderks, line 9: Where does 200px come from?

src/app/components/ImageViewer/ImageViewer.tsx
jonkoops: This was never working as the `download` attribute was set on a `button` element. Also the download is handled by a separate method passed in with a prop that does not support passing a file name.

src/app/containers/ActualityContainer.jsx
jonkoops: `<LoadingIndicator style={{ top: '200px' }} />` this original code

PR #2977, opened 2020-07-21
Use the history object from redux-first-router to handle url state ch…

royderks:
…anges


src/store/params-registry/paramRegistry.js
```diff
@@ -309,6 +310,63 @@ class ParamsRegistry {
...
+   * @param name
+   * @param value
+   * @param replace
+   * @returns void
+   */
+  setParam(param, value, replace = false) {
+    if (typeof window === 'undefined') {
+      return
+    }
+    const url = new URL(window.location.href)
```
chrisvanmook, line 324: Seems like URL is not supported in IE11 https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

src/store/params-registry/paramRegistry.js
```diff
@@ -309,6 +310,63 @@ class ParamsRegistry {
...
+    this.historyObject[action](url.pathname + url.search)
+  }
+
+  /**
+   * Get the requested param from the history object and decode it
+   * @param param
+   * @returns void
+   */
+  getParam(param) {
+    const value = new URLSearchParams(this.historyObject.location.search).get(param)
```
chrisvanmook, line 359: Also this one, no IE support https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

PR #2978, opened 2020-07-21
Changed the url for the parking layers geosearch, deleted redundant code

royderks:


src/shared/services/search/geosearch.js
```diff
@@ -159,10 +158,7 @@ export default function geosearch(location, user) {
         searchParams,
         false,
         true,
-      ).then(
-        (data) => ({ features: getFeaturesFromResult(endpoint.uri, data) }),
-        () => ({ features: [] }),
-      ) // empty features on failure of api call
+      ).then(() => ({ features: [] })) // empty features on failure of api call
```
jonkoops, line 161: This should be `.catch()`

PR #3003, opened 2020-07-28
Log environment information to console

jonkoops:
Also pass in the `COMMIT_HASH` when building a container in Jenkins

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 # Build
 FROM node:12 AS build-deps
+ARG GIT_COMMIT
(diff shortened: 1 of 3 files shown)


src/app/AppBody.jsx
```diff
@@ -75,10 +87,34 @@ const AppBody = ({
   const { enableLinkTracking } = useMatomo()
   enableLinkTracking()
 
+  const ie11ContentID =
```
chrisvanmook, line 90: perhaps rebase on develop?

PR #3005, opened 2020-07-29
Improve the hook that looks for changes in the map state or URLL

chrisvanmook:


src/app/pages/MapPage/Components/DrawTool.tsx
```diff
@@ -118,6 +119,20 @@ const DrawTool: React.FC<Props> = ({ onToggle, isOpen, setCurrentOverlay }) => {
...
+ * The bounds are not available on the `e` object and are already updated on the feature group (e.g. drawnItemsGroup.getLayers)
+ * 
+ * If found, they can be deleted like this:
+ * 
+ * if (existingDrawingGeometry) {
+      deleteDrawingGeometry(existingDrawingGeometry[0][0])
+    }
+ * 
+ */
+
```
alimpens, line 135: Can't you use the getBounds method on the layer here?

PR #3043, opened 2020-08-04
Override focus color to increase contrast on background elements

jonkoops:


jonkoops:
Closed this as we decided to use the browser default focus styling for accessibility reasons.

PR #3083, opened 2020-08-10
Fixed type error when panoPreview is not available

chrisvanmook:


src/environment.ts
```diff
@@ -5,7 +5,7 @@ const defaultEnvironment = {
   IIIF_ROOT: 'https://acc.images.data.amsterdam.nl/',
   API_ROOT: 'https://acc.api.data.amsterdam.nl/',
   CMS_ROOT: 'https://acc.cms.data.amsterdam.nl/',
-  GRAPHQL_ENDPOINT: 'https://acc.api.data.amsterdam.nl/cms_search/graphql/',
+  GRAPHQL_ENDPOINT: 'http://localhost:8080/cms_search/graphql/',
```
jonkoops, line 8: Uhmm... Doesn't look like this should have been merged @chrisvanmook

PR #3230, opened 2020-09-06
Added spinner to download button on Article

chrisvanmook:


src/app/components/EditorialPage/EditorialPage.jsx
```diff
@@ -36,6 +38,12 @@ const EditorialPage = ({
     }
   }, [documentTitle])
 
+  React.useEffect(() => {
+    if (error) {
+      window.location.replace(routing.niet_gevonden.path)
```
jonkoops, line 43: Instead of redirecting to the 404 page would it not make more sense to show an error message on the page itself?

jonkoops (approved):

PR #3269, opened 2020-09-09
Replace 'zoom in' messages in layers panel with an icon

jonkoops:
- Replaces the 'zoom in' message below each map layer in the layers panel with an icon
- Also shows the legend items of map layers regardless of zoom level

image


src/map/components/legend/MapLegend.jsx
```diff
@@ -341,72 +362,67 @@ const MapLegend = ({
...
+                          ? mapLayer.legendItems.map((legendItem, legendItemIndex) => {
+                              const legendItemIsVisible = legendItem.isVisible
+                              const LegendLabel = !legendItem.notSelectable
+                                ? StyledLabel
+                                : NonSelectableLegendParagraph
+                              return !legendItemIsVisible && printMode ? null : (
+                                <li
+                                  className="map-legend__item"
+                                  // eslint-disable-next-line react/no-array-index-key
+                                  key={legendItemIndex}
```
chrisvanmook, line 377: I know this was here already, but there is no way we can get a unique id here?

jonkoops: Actually yes, since we no longer have two elements here we can simply use the legend item id. Changed it.

chrisvanmook (approved):

PR #3331, opened 2020-09-22
Panorama feedback

jonkoops:
- Add tests for encodeParam
- Remove checks for result of useMapInstance
- Replace individual lodash packages with a single one
- Replace useMapClick with generic useMapEvents hook from ARM
- Some random code style improvements
- Optimize SVGs
- Random code cleanup


package.json
```diff
@@ -67,13 +67,7 @@
...
     "leaflet.nontiledlayer": "1.0.8",
     "lint-staged": "10.4.0",
-    "lodash.debounce": "4.0.8",
-    "lodash.defer": "4.1.0",
-    "lodash.get": "4.4.2",
-    "lodash.identity": "3.0.0",
-    "lodash.isequal": "4.5.0",
-    "lodash.isundefined": "3.0.1",
-    "lodash.throttle": "4.1.1",
+    "lodash": "^4.17.20",
```
janjaap, line 70: This change will increase the bundle size. Did you check the bundles after this change? Lodash adds ~500 KB of scripts

chrisvanmook: Indeed. You can use `lodash-es` however

PR #3359, opened 2020-10-01
Remove custom styling to restore mobile menu toggle

jonkoops:


src/app/components/Header/HeaderMenu.jsx
```diff
@@ -24,9 +24,14 @@ const StyledMenuInline = styled(MenuInline)`
     tall ? themeColor('tint', 'level2')({ theme }) : themeColor('tint', 'level1')({ theme })};
 `
 
+// TODO: Fix z-index issue with overlay in ASC eventually
```
janjaap, line 27: Is this another bug ticket?

src/app/components/Header/HeaderMenu.jsx
```diff
@@ -24,9 +24,14 @@ const StyledMenuInline = styled(MenuInline)`
...
+const StyledMenuToggle = styled(MenuToggle)`
+  z-index: 24;
```
janjaap, line 29: Why a magic number here?

src/app/components/SearchBar/SearchBar.tsx
```diff
@@ -16,13 +16,17 @@ import SearchBarFilter from '../SearchBarFilter'
...
+// TODO: Fix z-index issue with overlay in ASC eventually
 const StyledSearchBarToggle = styled(SearchBarToggle)`
   z-index: 23;
   position: absolute;
   left: 0;
   right: 0;
   top: 0;
+
+  & > button {
+    margin-right: 50px;
```
janjaap, line 28: Shouldn't this be `${themeSpacing()}` instead of a fixed pixel value?

jonkoops: This is the size of the button next to this, 50px is not dividable by our themespacing unit so in this case not.

src/app/components/Header/HeaderMenu.jsx
jonkoops: Since it needs to be higher than the other magic number, this is a quick fix on top of another hack. This code should be revised in the future when we rebuild the header.

src/app/components/Header/HeaderMenu.jsx
jonkoops: Yeah, I copied @chrisvanmook's original message. We'll have to resolve this when we rebuild the header.

chrisvanmook (approved):

src/app/components/Header/HeaderMenu.jsx
janjaap: Is that bug ticket already on the board? If not, can you create it, before we forget?

janjaap (approved):

src/app/components/Header/HeaderMenu.jsx
jonkoops: https://datapunt.atlassian.net/browse/DI-1164

PR #3366, opened 2020-10-02
Refactor detail pages to use route params

jonkoops:
Also reduces usage of Redux in various components.


src/app/pages/ConstructionFilesPage/ConstructionFilesPage.tsx
```diff
@@ -27,7 +27,12 @@ const StyledRow = styled(Row)`
 const ERROR_MESSAGE =
   'Er kunnen door een technische storing helaas geen bouw- en omgevingsdossiers worden getoond. Probeer het later nog eens.'
 
-const ConstructionFilesContainer = () => {
+interface ConstructionFilesPageParams {
+  id: string
+}
```
janjaap, line 32: This same signature is used in `CollectionDetailPage`. Wouldn't it be better to put this interface in a shared location?

src/api/metadata/getMetadata.test.ts
```diff
@@ -0,0 +1,21 @@
...
+describe('getMetadata', () => {
+  const response = [{ foo: 'bar' }]
+
+  it('makes a request and returns the response', async () => {
+    mockedFetchWithoutToken.mockReturnValueOnce(Promise.resolve(response))
+
+    await expect(getMetadata()).resolves.toEqual(response)
+
+    expect(mockedFetchWithoutToken).toHaveBeenCalledWith(
+      'https://acc.api.data.amsterdam.nl/metadata',
```
janjaap, line 18: In the actual file, you're using `environment.API_ROOT`. Why not use that here as well? It would make the test more generic in case of changes to the configuration.

src/app/pages/PublicationDetailPage/PublicationDetailPage.jsx
```diff
@@ -22,12 +21,13 @@ import getImageFromCms from '../../utils/getImageFromCms'
 import useDownload from '../../utils/useDownload'
 import useFromCMS from '../../utils/useFromCMS'
 
-const PublicationDetailPage = ({ id }) => {
+const PublicationDetailPage = () => {
+  const { id } = useParams()
```
janjaap, line 25: Other pages have a reference to an interface here. Why not use that reference here as well?

src/app/pages/SpecialDetailPage/SpecialDetailPage.jsx
```diff
@@ -1,20 +1,20 @@
...
-import { getLocationPayload } from '../../../store/redux-first-router/selectors'
 import ContentContainer from '../../components/ContentContainer/ContentContainer'
 import EditorialPage from '../../components/EditorialPage/EditorialPage'
 import IFrame from '../../components/IFrame/IFrame'
 import useFromCMS from '../../utils/useFromCMS'
 import Animation from './specials/Animation'
 
-const SpecialDetailPage = ({ id }) => {
+const SpecialDetailPage = () => {
+  const { id } = useParams()
```
janjaap, line 14: Same here; why not refer to the interface also used in other components?

src/app/routes.ts
```diff
@@ -14,7 +14,16 @@ export const MAIN_PATHS = {
   MAP: 'kaart',
 }
 
-export const routing = {
+// TODO: Refactor this file once redux-first-router has been removed from the project.
```
janjaap, line 17: Is there a ticket in the backlog that describes the removal of `redux-first-router` and does that ticket mention the refactor of this file? If not, can you either create the ticket or mention the refactor in that ticket?

src/app/utils/toSlug.test.ts
```diff
@@ -0,0 +1,12 @@
+import toSlug from './toSlug'
+
+describe('toSlug', () => {
+  it('should return a sluggified version from a string', () => {
+    expect(toSlug('A title ? With (special) 0 characters ! - #')).toBe(
+      'a-title-with-special-0-characters',
+    )
+
+    expect(toSlug('-A string with --- multiple -- slashes-')).toBe('a-string-with-multiple-slashes')
```
janjaap, line 9: Dashes. Not slashes.

src/header/components/auto-suggest/AutoSuggestItem.tsx
```diff
@@ -54,7 +54,7 @@ const AutoSuggestItem: React.FC<AutoSuggestItemProps> = ({
   const to: LocationDescriptorObject = useMemo(() => {
     if (suggestion.type === SearchType.Dataset) {
       const [, , id] = extractIdEndpoint(suggestion.uri)
```
janjaap, line 56: Isn't it possible to use `useParams` here as well?

src/header/components/auto-suggest/AutoSuggestItem.tsx
```diff
@@ -72,7 +72,7 @@ const AutoSuggestItem: React.FC<AutoSuggestItemProps> = ({
       suggestion.type === CmsType.Special
     ) {
       const [, , id] = extractIdEndpoint(suggestion.uri)
```
janjaap, line 74: And here

src/app/pages/PublicationDetailPage/PublicationDetailPage.jsx
jonkoops: This file is not written in TypeScript, refactoring it now would be a bit out of scope of the work in this PR. That said, we do intend to slowly convert all code to TypeScript.

src/app/pages/SpecialDetailPage/SpecialDetailPage.jsx
jonkoops: Also, out of scope see the previous comment.

src/app/routes.ts
jonkoops: There is a ticket in our backlog that describes the removal of redux-first-router. It does not describe this file specifically, but considering the direct dependency from others (e.g. the `redux-first-router` reducer) it should not be hard to find. I am counting on the ingenuity of the dev here.

src/api/metadata/getMetadata.ts
```diff
@@ -0,0 +1,23 @@
...
+  last_import_date: string | null
+  /* eslint-enable camelcase */
+}
+
+/**
+ * Retrieve metadata about data sources used on the site, such as the update frequency and the last time a data source was update.
+ *
+ * API documentation: https://api.data.amsterdam.nl/metadata/
+ */
+export async function getMetadata() {
```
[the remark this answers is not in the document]
jonkoops: No particular reason. Sometimes I prefer using the `function` keyword when exporting from modules, since it's possible to do a default export on them without seperating the definition. For example:

```ts
export default async function getMetadata() {
  return fetchWithoutToken<Metadata[]>(joinUrl(environment.API_ROOT, 'metadata'))
}
```

src/app/pages/ActualityPage/ActualityPage.tsx
```diff
@@ -0,0 +1,77 @@
...
+import usePromise, { PromiseStatus } from '../../utils/usePromise'
+
+const StyledLoadingSpinner = styled(LoadingSpinner)`
+  position: absolute;
+  top: 200px;
+`
+
+const ActualityPage: FunctionComponent = () => (
+  <div className="c-page">
+    <div
```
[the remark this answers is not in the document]
jonkoops: Agreed, however I would safe this optimization for now. At some point in the future we will eliminate all SASS styles, so this component will also have to be refactored.

src/api/metadata/getMetadata.test.ts
jonkoops: Sound's good to me, I've updated this code.

src/app/utils/toSlug.test.ts
jonkoops: Fixed

src/header/components/auto-suggest/AutoSuggestItem.tsx
jonkoops: We cannot use hooks conditionally since React relies on their order of execution. In this case we're also passing in a URL which is not necessarily the same as the current route, which is what is used by React Router's `useParams`.

src/header/components/auto-suggest/AutoSuggestItem.tsx
jonkoops: Same here :)

src/app/pages/ConstructionFilesPage/ConstructionFilesPage.tsx
jonkoops: Not all detail pages have the same structure for params so I would like to keep this next to the component and in the future possibly next to the actual definition of the route.

chrisvanmook (approved):

src/app/pages/ArticleDetailPage/ArticleDetailPage.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+        } as any),
+    )
+
+    mount(
+      <ThemeProvider>
+        <ArticleDetailPage />
+      </ThemeProvider>,
+    )
+
+    expect(useFromCMS).toHaveBeenCalledWith(cmsConfig.ARTICLE, 'foo')
```
janjaap, line 71: Is it important to know how the component gets its data? Or is it more important to check what has been rendered after the data has successfully been retrieved? Personally, I would say the latter. Quite a big part of the `ArticleDetailPage` component isn't covered by test cases that assert that specific sections on the page are rendered

jonkoops: In this case we are asserting that when the component is mounted it will correctly pass the id of the object.

jonkoops: As for the test coverage, this is something we will need to improve. In this case we're just ensuring the existing tests keep working.

PR #3386, opened 2020-10-06
Reduce usage of Lodash

jonkoops:


src/map/containers/draw-tool/DrawToolContainer.jsx
```diff
@@ -112,9 +111,7 @@ class DrawToolContainer extends React.Component {
 
   getMarkers() {
     const { geometry, dataSelection } = this.props
-    return geometry && geometry.length > 0
-      ? geometry
-      : get(dataSelection, 'geometryFilter.markers', [])
+    return geometry && geometry.length > 0 ? geometry : dataSelection?.geometryFilter?.markers ?? []
```
janjaap, line 114: Or
```
return geometry?.length > 0 ? geometry : dataSelection?.geometryFilter?.markers ?? []
```

src/shared/services/data-selection/data-selection-api-brk.js
```diff
@@ -49,7 +48,7 @@ export function getMarkers(config, activeFilters, zoomLevel, boundingBox) {
             id: generateId(),
             type: 'dataSelectionBounds',
           },
-        ].filter(identity),
+        ].filter((item) => item),
```
janjaap, line 51: Or `.filter(Boolean)`

src/shared/services/data-selection/data-selection-api.js
```diff
@@ -52,7 +51,7 @@ function recurGetContent(path, rawData) {
     const key = path[0]
     const rawValue = rawData[key]
 
-    return Array.isArray(rawValue) ? rawValue.filter(identity).join(' | ') : rawValue
+    return Array.isArray(rawValue) ? rawValue.filter((part) => part).join(' | ') : rawValue
```
janjaap, line 54: Or `.filter(Boolean)`

src/shared/services/document-title/document-title.js
```diff
@@ -43,8 +43,9 @@ export const detailDocumentTitleWithName = (action) => {
   // We fill the title for details in 2 steps
   let title = document.title.replace(' - Data en informatie - Amsterdam', '')
 
-  const isDataset = !!get(action, 'payload.data.editDatasetId', null)
-  if (isDataset && title.indexOf(':') === -1) {
+  const dataSetId = action?.payload?.data?.editDatasetId ?? null
+
+  if (dataSetId !== null && title.indexOf(':') === -1) {
```
janjaap, line 48: Is it necessary to check explicitly for `null`? `dataSetId` Will evaluate to false anyway if it's null.
So, `if (dataSetId && title.indexOf(':') === -1) {` would do same.

src/shared/services/search/geosearch.js
```diff
@@ -121,7 +120,7 @@ function getRelatedObjects(geosearchResults, user) {
 
         Promise.all(requests).then((results) => {
           const geosearchResultsCopy = [...geosearchResults]
-          const filteredResults = results.filter(identity)
+          const filteredResults = results.filter((result) => result)
```
janjaap, line 123: Or `.filter(Boolean)`

chrisvanmook (approved):

src/shared/services/document-title/document-title.js
jonkoops: I don't know if `dataSetId` can be a number, if so it could be a `0` which might evaluate to a falsy condition unwarranted.

src/shared/services/data-selection/data-selection-api-brk.js
jonkoops: This gets typecast under the hood anyways, the result ends up the same. In this case I don't think adding a boolean constructor adds anything.

src/map/containers/draw-tool/DrawToolContainer.jsx
jonkoops: Good catch, added this change.

PR #3397, opened 2020-10-07
(Chore) Debounce instead of throttle pano orientation change

janjaap:
This PR replaces the panoramo orientation change event action trigger with a debounce function. The throttle function fired every 300 ms, but fired a lot of times when rotating a panorama. IMHO, it's better for performance to only update the pano orientation in the state every 300 ms when orientation is still in progress.

Note that, in order to make the test for the `useMarzipano` hook work, the `canvas` dependency has been added.


chrisvanmook:
This is the "old" panorama container. For the new panoramaviewer we use `useMarzipano.ts` where this throttle is also used, could you update it there as well?

src/panorama/containers/PanoramaContainer.jsx
```diff
@@ -1,5 +1,5 @@
 import classNames from 'classnames'
-import throttle from 'lodash.throttle'
```
jonkoops, line 2: This `throttle` is also used in `useMarzipano.ts`, could you replace it there as well? I think we can also get rid of the `lodash.throttle` and `@types/lodash.throttle` dependencies as well.

jonkoops (changes requested):

package.json
```diff
@@ -151,6 +149,7 @@
     "babel-loader": "^8.1.0",
     "babel-plugin-styled-components": "^1.11.1",
     "babel-plugin-transform-commonjs-es2015-modules": "^4.0.1",
+    "canvas": "^2.6.1",
```
chrisvanmook, line 152: why the canvas dependency?

src/app/utils/useMarzipano.test.ts
```diff
@@ -0,0 +1,22 @@
...
+describe('useMarzipano', () => {
+  it('initialises the Marzipano Viewer', () => {
+    const ref = { current: null }
+
+    const { result } = renderHook(() => useMarzipano(ref))
+
+    expect(result.current.marzipanoViewer).toBeNull()
+
+    const div = document.createElement('div')
+    document.body.appendChild(div)
```
jonkoops, line 14: Does the element actually need to be in the body?

src/panorama/containers/PanoramaContainer.jsx
```diff
@@ -40,7 +40,7 @@ class PanoramaContainer extends React.Component {
     this.hotspotClickHandler = this.hotspotClickHandler.bind(this)
     this.loadPanoramaScene = this.loadPanoramaScene.bind(this)
 
-    this.updateOrientationThrottled = throttle(this.updateOrientation, 300, {
+    this.updateOrientationThrottled = debounce(this.updateOrientation, 300, {
```
jonkoops, line 43:
```suggestion
    this.updateOrientationDebounced = debounce(this.updateOrientation, 300, {
```

src/app/utils/useMarzipano.ts
```diff
@@ -19,8 +25,9 @@ const useMarzipano = (ref: RefObject<HTMLElement>) => {
     if (heading && pitch && fov) {
       setCurrentMarzipanoView({ heading, pitch, fov })
     }
-  }
-  const updateOrientationThrottled = throttle(updateOrientation, 300, {
+  }, [])
+
+  const updateOrientationThrottled = debounce(updateOrientation, 300, {
```
jonkoops, line 30:
```suggestion
  const updateOrientationDebounced = debounce(updateOrientation, 300, {
```

src/app/utils/useMarzipano.ts
```diff
@@ -1,15 +1,21 @@
...
 import { getOrientation } from '../../panorama/services/marzipano/marzipano'
 import { PanoParam } from '../pages/MapPage/query-params'
 
-const useMarzipano = (ref: RefObject<HTMLElement>) => {
+const useMarzipano = (
+  ref: RefObject<HTMLElement>,
+): {
+  marzipanoViewer: any
+  marzipanoViewerRef: RefObject<HTMLElement>
+  currentMarzipanoView: any
```
jonkoops, line 13: According to the types defined in this method `currentMarzipanoView` is of type `PanoParam | null`. This type should be updated to reflect this.

```suggestion
  currentMarzipanoView: PanoParam | null
```

jonkoops, line 13: According to the type definitions used in this method the type of `currentMarzipanoView` should be `PanoParam | null`.

```suggestion
  currentMarzipanoView: PanoParam | null
```

That said the `PanoParam` interface should more correctly be named `Pano` as it does not implement the `UrlParam` interface and thus is not actually a param, but rather the value produced from it.

src/app/utils/useMarzipano.ts
```diff
@@ -1,15 +1,21 @@
...
+} => {
   const [marzipanoViewer, setMarzipanoInstance, marzipanoViewerRef] = useStateRef<any>(null)
   const [currentMarzipanoView, setCurrentMarzipanoView] = useState<PanoParam | null>(null)
 
-  const updateOrientation = () => {
+  const updateOrientation = useCallback(() => {
```
jonkoops, line 18: Is this change really needed? According to the [documentation](https://reactjs.org/docs/hooks-reference.html#usecallback) of React:

> useCallback will return a memoized version of the callback that only changes if one of the dependencies has changed. This is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders (e.g. shouldComponentUpdate).

This is not the case here, we are simply defining a function that later on is registered only once as an event listener on a 3rd party library. Memoizing this method seems unnecessary.

jonkoops (changes requested):

package.json
janjaap: I added canvas, because Jest was complaining about functionality that depended on it. In the end, I wasn't able to test the `useMarzipano` hook in full and decided to scrap some tests, but forgot to remove the `canvas` dependency. Good catch, thanks. I'll remove it.

src/app/utils/useMarzipano.test.ts
janjaap: No, it doesn't. I'll remove the line.

src/app/utils/useMarzipano.ts
janjaap: That's the reason I wrapped it in `useCallback`, so that the function never changes. Now, with every component render, it will generate a new function and place it in memory. Chrome deals pretty well with garbage collecting unreferenced functions, but other browsers, specifically Firefox, don't and the memory consumption of the browser will increase over time.

package.json
chrisvanmook: Still not removed?

src/app/utils/useMarzipano.ts
jonkoops: Hmmm... I would not be optimizing for JS engines here as the best practices of today are the bad practices for tomorrow. It could very well be that wrapping things in this `useCallback` makes it harder for the JS engine to analyze the code and inline this method if it gets hot. For example Firefox is enabling a [new version](https://www.phoronix.com/scan.php?page=news_item&px=Firefox-Nightly-JS-Warp) of their JIT soon, which might change this situation.

Due to the behaviour of `useCallback` it will most likely allocate more memory. If this specific instance had any dependencies it would also keep those dependencies and the callback in memory even after they have changed. Kent Dodds [article](https://kentcdodds.com/blog/usememo-and-usecallback) explains this pretty well, of which the TLDR is:

> ## So when should I useMemo and useCallback?
> There are specific reasons both of these hooks are built-into React:
> - Referential equality
> - Computationally expensive calculations

Since this code is neither I would still recommend to remove `useCallback` in this instance.

jonkoops (approved):

src/app/utils/useMarzipano.ts
janjaap: When applying a `useCallback` handler, I usually test the amount of functions that are generated at run-time by the component. Like so:
```jsx
const functions = new Set()

const MyComponent = () => {
  const internalFunction = () => {}
  functions.add(internalFunction)

  return (
    <span>num functions: {functions.size}</span>
  )
}
```
Not using `useCallback` in this case will generate ~10 functions whereas wrapping the internal function with `useCallback` will only create one instance of said function in memory.

jonkoops: It will generate functions yes, but JavaScript engines are optimized to handle these kind of cases. The garbage collector will just throw them away as long as the function is not referenced anywhere.

In my opinion we should let the browser engine what it does best and just use these methods for computationally intensive tasks or references that should not be changed due to any kind of dependency tracker as recommended by the React documentation and various other resources.

That said if we really care that much about how often this method is allocated to memory it would make a lot more sense to simply put this method inside of the `useEffect` as follows (untested):


```ts
useEffect(() => {
  if (!ref.current) {
    return undefined
  }

  const viewer = new Marzipano.Viewer(ref.current, {
    // Set a stage type that is supported in the test environment
    stageType: process.env.NODE_ENV === 'test' ? 'css' : 'webgl',
    stage: {
      preserveDrawingBuffer: true,
      width: 960,
    },
  })

  const updateOrientation = () => {
    const { heading, pitch, fov } = getOrientation(viewer)

    if (heading && pitch && fov) {
      setCurrentMarzipanoView({ heading, pitch, fov })
    }
  }

  const updateOrientationDebounced = debounce(updateOrientation, 300, {
    leading: true,
    trailing: true,
  })

  setMarzipanoInstance(viewer)

  viewer.addEventListener('viewChange', updateOrientationDebounced)

  return () => {
    if (marzipanoViewerRef.current) {
      setMarzipanoInstance(null)
    }
  }
}, [ref])
```

PR #3398, opened 2020-10-07
Add Dcatd API and refactor existing API code

jonkoops:


src/api/dcatd/datasets/getDatasetById.ts
```diff
@@ -0,0 +1,14 @@
...
+import environment from '../../../environment'
+import { fetchWithToken } from '../../../shared/services/api/api'
+import { DcatDataset } from './types'
+
+/**
+ * Get a data catalogue entry by its identifier.
+ *
+ * API documentation: https://api.data.amsterdam.nl/api/swagger/?url=/dcatd/openapi#/default/get_datasets__id_
+ */
+// eslint-disable-next-line import/prefer-default-export
```
janjaap, line 11: Why would you disable the linter here? The rule has been configured for a reason.

jonkoops: Since I use that named export directly passing it through `index.ts`. In this case I think it's fine to do so, we're also doing this in different places.

chrisvanmook (approved):

src/api/dcatd/datasets/getDatasetById.ts
janjaap: So, why do we have the lint rule in place, then?

PR #3400, opened 2020-10-08
Remove Redux from dataset detail page

jonkoops:


src/app/utils/joinUrl.ts
```diff
@@ -5,13 +5,13 @@ const PATH_SEPARATOR = '/'
...
  * ```ts
- * joinUrl('http://example.com', 'foo/bar', '10') // http://example.com/foo/bar/10
+ * joinUrl(['http://example.com', 'foo/bar', '10']) // http://example.com/foo/bar/10
  * ```
  *
  * @param paths The paths of the URL to combine.
  * @param trailing If the returned value should have a trailing slash
  */
-export default function joinUrl(paths: string[], trailing = true) {
+export default function joinUrl(paths: string[], trailing = false) {
```
chrisvanmook, line 14: why false by default?

src/root-saga.js
```diff
@@ -28,7 +27,6 @@ export default function* rootSaga() {
     fork(watchFetchNearestDetails),
     fork(watchAuthenticationRequest),
     fork(watchFetchDataSelection),
-    fork(watchFetchDatasets),
```
chrisvanmook, line 31: awww yeah 🎉

src/shared/services/date-formatter/date-formatter.js
```diff
@@ -5,18 +5,22 @@ import { DEFAULT_LOCALE } from '../../config/locale.config'
...
+  // TODO: Remove this check, TypeScript should take care of it.
+  if (!date) {
+    return null
+  }
+
+  const dateInstance = typeof date === 'string' ? new Date(date) : date
+
+  return dateInstance.toLocaleDateString(DEFAULT_LOCALE, {
+    ...(day && { day: 'numeric' }),
+    ...(month && { month: 'long' }),
```
chrisvanmook, line 21: On the detail pages we sometimes use '2-digits' or just numeric. Think we have 2 options here:
- Make this configurable by passing the options instead of day, month, year (`Intl.DateTimeFormatOptions`) 
- Perhaps we should not use this util and stay close to vanilla JS. IMHO this util only exists because we don't have TS everywhere yet

chrisvanmook (changes requested):

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
```diff
@@ -13,46 +8,41 @@ import {
...
-  return input ? input.toLowerCase().replace(/[: ][ ]*/g, '-') : ''
-}
-
-const optionLabel = (input: string, list: { id: string; label: string }[], namespace?: string) => {
-  // returns the label of an option from an option list
-  // the list array elements contain at least the id and label properties
-  const prefix = isDefined(namespace) ? `${namespace}:` : ''
-  const index = list && list.findIndex((item) => prefix + item.id === input)
-  return index > -1 ? list[index].label : input
+function kebabCase(input: string) {
```
janjaap, line 31: No need to set the type of the return value?

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
```diff
@@ -13,46 +8,41 @@ import {
...
-
-const capitalizeFirstLetter = (string: string) => {
-  return string.charAt(0).toUpperCase() + string.slice(1)
+/**
+ * Gets the label by the identifier of the specified options, if no option could be found it will default to the id.
+ *
+ * @param id The identifier of the option to match.
+ * @param options The options to get the label from.
+ */
+function getOptionLabel(id: string, options: DatasetFilterOption[]) {
```
janjaap, line 41: This function seems to return either an object or a string; wouldn't it be better to set the type of the return value as well?

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
```diff
@@ -13,46 +8,41 @@ import {
...
+  return options.find((item) => item.id === id)?.label ?? id
 }
 
-const getFileSize = (bytes: number) => {
+function getFileSize(bytes: number) {
```
janjaap, line 45: What is the reason for changing `const` into `function`? None of the functions in this component are using references to `this`, so there should be no need to change the function from anonymous to named.

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
```diff
@@ -81,19 +73,26 @@ const getFileSize = (bytes: number) => {
   return `${number.toLocaleString()} ${units[power]}`
 }
 
-const getTimePeroid = (input: { ['time:hasBeginning']: string } | string) => {
-  if (isObject(input)) {
-    const startDate = input['time:hasBeginning'] && new Date(input['time:hasBeginning'])
-    const endDate = input['time:hasEnd'] && new Date(input['time:hasEnd'])
-    const result = startDate ? `${formatDate(startDate)} ` : ''
+function getTimePeriodLabel(period: DcatTemporal) {
+  const startDate = period['time:hasBeginning'] ? new Date(period['time:hasBeginning']) : null
```
janjaap, line 77: Couldn't 
```javascript
new Date(period?.['time:hasBeginning']) ?? null
``` 
also be used here? The `Date` constructor should return `undefined` for an invalid constructor param.

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
```diff
@@ -102,352 +101,364 @@ const StyledCustomHTMLBlock = styled(CustomHTMLBlock)`
...
   }
 `
+
 const Content = styled.div`
   width: 100%;
 `
 
+interface DatasetDetailPageParams {
+  id: string
+  slug: string
```
janjaap, line 111: `slug` Doesn't seem to be used

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
```diff
@@ -102,352 +101,364 @@ const StyledCustomHTMLBlock = styled(CustomHTMLBlock)`
...
+                          className="dcatd-button--edit"
+                          onClick={() => redirectToDcatd(datasetId)}
+                        >
+                          Wijzigen
+                          <span className="u-sr-only">Wijzigen</span>
+                        </Button>
+                      </div>
+                    )}
+                  </h3>
+                  <h2 className="o-header__title u-margin__bottom--3 ">{dataset['dct:title']}</h2>
```
janjaap, line 176: This is not a valid HTML document structure; the `h2` should come before the `h3`. Looking at the element order in the page, the `h3` should not be a heading element, the `h2` should be an `h1`, the `h3`s should be `h2` and `h4` should be `h3`.

janjaap (changes requested):
A couple of remarks; the semantic issues in the `DatasetDetailPage` component are IMO blocking.

src/app/utils/joinUrl.ts
jonkoops: Most APIs want to end on a non-trailing slash, for example `datasets/1234` so it makes sense not to include it by default.

janjaap: I think that the general rule of thumb is that list endpoints end with a slash and detail endpoints do not. Not sure if that rule is written in stone, tho.

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
jonkoops: In this case TypeScript can infer the return type based on the value we return, which is always a string.

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
jonkoops: As far as I can tell this method will always return a string, see the inferred type:

image

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
jonkoops: No particular reason, in this case the method does not immediately return a value so I don't see a benefit in using an arrow function.

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
jonkoops: Running `new Date(undefined)` in my console (see below) I am getting back a constructed Date object which holds an invalid date. So unless I am missing something here I do not think this is possible.

image

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
jonkoops: Yeah, it's not used but it does exist here so I decided to add the type anyways. This will get compiled away in the final bundle as the types are eliminated.

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
janjaap: Is that a rule; that far arrow functions should return a value?

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
janjaap: You're right; I was mistaken.

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
jonkoops: Agreed, I fixed up the headings. I kept the 'subtitle' as a heading element as it is the heading for the text below it.

src/shared/services/date-formatter/date-formatter.js
jonkoops: Yeah, agreed. I've cleaned up the utility and related code and tests.

src/app/utils/joinUrl.ts
jonkoops: Some list endpoints also do not end with a slash, it's a bit of a mess for sure.

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
jonkoops: As in a linting rule? Not at this moment no. As far as I know we have no rules on this.

janjaap: Not necessarily a linting rule, more like a team agreement.

jonkoops: Not at this time, hopefully something we can agree on once we have space to fix up our linting.

src/app/pages/DatasetDetailPage/DatasetDetailPage.tsx
```diff
@@ -158,7 +158,10 @@ const DatasetDetailPage: FunctionComponent = () => {
             <Column span={{ small: 1, medium: 2, big: 6, large: 12, xLarge: 12 }}>
               <Content className="dataset-detail">
                 <div className="o-header">
-                  <h3 className="o-header__subtitle">
+                  <h1 className="o-header__title u-margin__top--0 u-margin__bottom--1">
```
janjaap, line 161: This changes the structure of the document; it's semantically valid, but now the text `Dataset` will be displayed below the `h1`, where it was placed above that element before.

jonkoops: Yeah, I checked this. We're not using it on any other page so in this case it is fine.

janjaap (approved):

package.json
```diff
@@ -69,7 +69,7 @@
     "lodash.debounce": "^4.0.8",
     "lodash.defer": "^4.1.0",
     "lodash.isequal": "^4.5.0",
-    "marked": "^1.2.0",
+    "lodash.throttle": "^4.1.1",
```
janjaap, line 72: Wasn't this package removed in a previous PR?

jonkoops: Good catch, must have missed it during the merge. Removed.

chrisvanmook (approved):

PR #3305, opened 2020-09-16
Autosuggest improvements

chrisvanmook:

PR #3405, opened 2020-10-09
[DI-1154] Link to addresses from search results

janjaap:
This PR fixes an issue where certain query parameters weren't persisted to the state. The bug was introduced in https://github.com/Amsterdam/atlas/pull/3305 with the removal of the `query` property from the search page state. Some of the changes of said PR are reverted in this PR which makes the functionality work again.


chrisvanmook (commented):
I was actually hoping we could get rid of query's attached to `queryParameters.js` by using our new `useParam` hook. But perhaps we do have to wait until we can remove redux-first-router as this will cause issues when using together...

src/app/pages/SearchPage/DataSearchResults.tsx
```diff
@@ -35,7 +51,7 @@ const DataSearchResults: React.FC<DataSearchResultsProps> = ({
   const Card = compact ? DataCard : DataList
 
   // Get the total count for all data types
-  const totalCount = results.length ? [0, ...results].reduce((acc, cur) => acc + cur.count) : 0
+  const totalCount = results.length ? results.reduce((acc: number, { count }) => acc + count, 0) : 0
```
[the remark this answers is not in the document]
janjaap: It is, yes. I changed it.

src/app/pages/SearchPage/DataSearchResults.tsx
```diff
@@ -15,11 +15,27 @@ const CardWrapper = styled.div<{ compact: boolean }>`
   margin-bottom: ${({ compact }) => (compact ? themeSpacing(2) : themeSpacing(8))};
 `
 
+type SingleResult = {
+  endpoint: String
```
jonkoops, line 19: It's better to use the lower-case `string` here to refer to the primitive type of string rather than the constructor type. For more information see the TypeScript [documentation](https://www.typescriptlang.org/docs/handbook/basic-types.html#about-number-string-boolean-symbol-and-object).

src/app/pages/SearchPage/DataSearchResults.tsx
```diff
@@ -35,7 +51,7 @@ const DataSearchResults: React.FC<DataSearchResultsProps> = ({
   const Card = compact ? DataCard : DataList
 
   // Get the total count for all data types
-  const totalCount = results.length ? [0, ...results].reduce((acc, cur) => acc + cur.count) : 0
+  const totalCount = results.length ? results.reduce((acc, { count }) => acc + count, 0) : 0
```
jonkoops, line 54: Nice cleanup! I think it could actually be reduced even more by removing the ternary entirely since a `reduce` on an empty array is a no-op. 

```suggestion
  const totalCount = results.reduce((acc, { count }) => acc + count, 0) 
```

src/app/pages/SearchPage/SearchPageDucks.ts
```diff
@@ -94,6 +96,7 @@ type StoreValue = { [REDUCER_KEY]: SearchPageState }
 export const getSort = ({ [REDUCER_KEY]: { sort } }: StoreValue) => sort
 export const getPage = ({ [REDUCER_KEY]: { page } }: StoreValue) => page
 
+export const getQuery = ({ [REDUCER_KEY]: { query } }: StoreValue) => query?.toString()
```
jonkoops, line 99: There is no need to use optional chaining here as the `SearchPageState` guarantees that `query` is always present and of type `string`.  If this is not the case the code that sets this value should be changed.

```suggestion
export const getQuery = ({ [REDUCER_KEY]: { query } }: StoreValue) => query
```

src/store/params-registry/paramRegistry.js
```diff
@@ -193,24 +193,13 @@ class ParamsRegistry {
   }
 
   setQueriesFromState(currentLocationType, state, nextAction) {
-    // Temporary fix to allow new params (using the useParam hook)
```
jonkoops, line 196: I don't have a clear indication if removing this code will break anything. @chrisvanmook perhaps you can judge this this better.

src/app/pages/SearchPage/DataSearchResults.tsx
```diff
@@ -15,11 +15,27 @@ const CardWrapper = styled.div<{ compact: boolean }>`
   margin-bottom: ${({ compact }) => (compact ? themeSpacing(2) : themeSpacing(8))};
 `
 
+type SingleResult = {
```
jonkoops, line 18: Personally I prefer to use `interface` over `type` in these cases as combining types can become messy very quickly especially in error messages from the TypeScript compiler. However we have been mixing these in the project already and since we don't have any linting in place the point is moot.

src/app/pages/SearchPage/DataSearchResults.tsx
```diff
@@ -15,11 +15,27 @@ const CardWrapper = styled.div<{ compact: boolean }>`
...
+  type: string
+  __typename: 'CombinedDataResult'
+}
+
 export interface DataSearchResultsProps {
   page: string
   query: string
-  // TODO: Properly type the results
-  results: any[]
+  results: Array<DataSearchResults>
```
jonkoops, line 38: An alternative syntax you could use for arrays is `DataSearchResults[]`, but both are equally valid.

jonkoops (changes requested):

src/app/pages/SearchPage/DataSearchResults.tsx
janjaap: Could do that, but it will break when `results` is anything but an array.

src/store/params-registry/paramRegistry.js
janjaap: This particular change reverts a previous change by @chrisvanmook

src/app/pages/SearchPage/DataSearchResults.tsx
jonkoops: In this case we're relying on TypeScript to guarantee that this value is never anything else than an array (see the `DataSearchResults` interface). So checking it again here at runtime is not needed, unless of course the response can contain other values. In that case the type should be updated to reflect that.

src/app/pages/SearchPage/DataSearchResults.tsx
```diff
@@ -15,11 +15,27 @@ const CardWrapper = styled.div<{ compact: boolean }>`
   margin-bottom: ${({ compact }) => (compact ? themeSpacing(2) : themeSpacing(8))};
 `
 
+interface SingleResult {
```
jonkoops, line 18: Perhaps ` DataResult` would be be a more appropriate name, like the `__typename` from GQL.

src/app/pages/SearchPage/DataSearchResults.tsx
```diff
@@ -15,11 +15,27 @@ const CardWrapper = styled.div<{ compact: boolean }>`
...
+  endpoint: string
+  id: string
+  label: string
+  subtype: string
+  type: string
+  __typename: 'DataResult'
+}
+
+interface DataSearchResults {
```
jonkoops, line 27: Perhaps ` CombinedDataResult` would be be a more appropriate name, like the `__typename` from GQL.

jonkoops (approved):
Looks good to me if @chrisvanmook can give his judgement as well.

PR #3407, opened 2020-10-09
[DI-1190] Buurten links fix

janjaap:
This PR fixes an issue where an error would be thrown whenever a link was clicked for which the Matomo data required a prop `modus` to be present in the `query` object. Since the check for that prop was too rigid, not taking into account that `query` could also be not provided, error were thrown and navigation was prevented.

In addition to the fix, this PR also  renames the prop `firstAction` to `isFirstAction`, since it's a boolean value. Also, a deeply nested ternary condition inside the `trackEvents` script has been replaced with a more readable construct. In order to ensure that the replacement didn't break anything, tests were written prior to refactoring.

### DIFF
--- a/src/store/middleware/matomo/trackEvents.js
+++ b/src/store/middleware/matomo/trackEvents.js
@@ -43,45 +43,60 @@ import {
   isPanoPage,
 } from '../../redux-first-router/selectors'
 
+export const TRACK_ACTION_NAVIGATION = 'navigation'
+export const TRACK_CATEGORY_AUTO_SUGGEST = 'auto-suggest'
+export const TRACK_NAME_ENLARGE_DETAIL_MAP = 'detail-kaart-vergroten'
+export const TRACK_NAME_FULL_DETAIL = 'detail-volledig-weergeven'
+export const TRACK_NAME_LEAVE_PANO = 'panorama-verlaten'
+
 /* istanbul ignore next */
 const trackEvents = {
   // NAVIGATION
   // NAVIGATION -> NAVIGATE TO DATA DETAIL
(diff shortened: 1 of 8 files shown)


src/store/middleware/matomo/trackEvents.js
```diff
@@ -43,45 +43,60 @@ import {
...
-        ]
-      : isPanoPage(state)
-      ? [
-          MATOMO_CONSTANTS.TRACK_EVENT,
-          'navigation', // NAVIGATION -> CLICK CLOSE FROM PANORAMA
-          'panorama-verlaten',
-          null,
-        ]
-      : []
+  [routing.dataDetail.type]: function trackDataDetail({ isFirstAction, query, tracking, state }) {
```
jonkoops, line 56: Nice! This is a lot easier on the eyes to read than the tenary version.

src/store/middleware/matomo/trackEvents.test.js
```diff
@@ -0,0 +1,93 @@
...
+  TRACK_NAME_FULL_DETAIL,
+  TRACK_NAME_LEAVE_PANO,
+} from './trackEvents'
+
+describe('trackEvents', () => {
+  it('matches the snapshot', () => {
+    // snapshots generally don't offer much value, but in this case, since the value of trackEvents is an object with
+    // keys that are specific to the sections in the application that need to be tracked, it's worthwhile to generate
+    // a snaphot of the entire object so that any changes will be more easily picked up
+    expect(trackEvents).toMatchSnapshot()
```
jonkoops, line 20: This makes sense, in the future TypeScript would give us these guarantees when compiling. In the near future we hope to get rid of this tracking method all together.

jonkoops (approved):
LGTM :+1:

PR #3423, opened 2020-10-12
(Chore) GitHub actions adjustments

janjaap:
This PR:
- Makes sure that Github actions are run on all PRs, regardless of base branch
- Splits the action into multiple, separate, actions. It runs the actions in parallel and it allows us to see which individual action failed and will show up like so in a PR:
Screenshot 2020-10-12 at 07 58 27

- Generates a separate artifact for the test coverage

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,10 +47,3 @@ jobs:
 
       - name: Run ${{ matrix.job }}
         run: npm run ${{ matrix.job }}
-
-      - name: Upload coverage report
-        uses: actions/upload-artifact@v1
-        if: matrix.job == 'test'
-        with:
-          name: coverage-report
-          path: coverage/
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -37,11 +36,14 @@ jobs:
             ${{ runner.os }}-build-${{ env.cache-name }}-
             ${{ runner.os }}-build-
             ${{ runner.os }}-
+
+      - name: Setup Node 12
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.18
```
jonkoops, line 43: Why use a specific version of NodeJS here instead of the latest LTS version as specified in the matrix? Also do we even need to set up node again here or is the earlier command still valid?

jonkoops (changes requested):

.github/workflows/ci.yml
janjaap: Good catch. I copied it from `signals-frontend` and forgot to remove this step

jonkoops (approved):

chrisvanmook (approved):

PR #3428, opened 2020-10-12
Major Refactor: Moved detail pages from angularJS to React.

chrisvanmook:
In this PR I've:
- Copied fields from detail templates in AngularJS to the `map-services.config.ts` file.
- Added some normalisations for some detail types
- Added new Detail-related components  (LinkList, GroupedItems etc.)
- Removed detail-saga's. This is now moved to the temporary `useDataDetail` hook
- Refactored code when possible

Best way to test is comparing pages on acceptance to this branch and check if all the data is present.
Also, since we don't get any typings from the API's, it might be possible that some info won't show up when it should or vice versa. This is something we need to take in account when updating our E2E tests


src/app/pages/DataDetailPage/DataDetailPage.tsx
```diff
@@ -67,97 +36,30 @@ const TypeHeading = styled(DetailHeading)`
...
-            <PanelContents legacyLayout result={result} />
-          </DetailWrapper>
-        </DetailContainer>
+    <PromiseResult<any> promise={promise} onRetry={onRetry}>
+      {({ result }) => (
+        <DetailWrapper>
+          <TypeHeading>{result.value?.data.title}</TypeHeading>
+          <HeadingWrapper>
+            <Heading as="h1">{getPanelTitle(result)}</Heading>
+            {!!result?.value?.data?.infoBox && <DetailInfoBox {...result?.value?.data?.infoBox} />}
```
janjaap, line 56: Why is the double bang necessary here?

src/app/pages/DataDetailPage/DataDetailPage.tsx
```diff
@@ -67,97 +36,30 @@ const TypeHeading = styled(DetailHeading)`
...
-            </HeadingWrapper>
-            <PanelContents legacyLayout result={result} />
-          </DetailWrapper>
-        </DetailContainer>
+    <PromiseResult<any> promise={promise} onRetry={onRetry}>
+      {({ result }) => (
+        <DetailWrapper>
+          <TypeHeading>{result.value?.data.title}</TypeHeading>
+          <HeadingWrapper>
+            <Heading as="h1">{getPanelTitle(result)}</Heading>
```
janjaap, line 55: Isn't `Heading` an `h1`, unless specified otherwise?

src/app/pages/MapPage/detail/DetailDefinitionList.tsx
```diff
@@ -1,23 +1,34 @@
...
+const StyledCustomHTMLBlock = styled(CustomHTMLBlock)`
+  white-space: pre-line;
+`
+
 const DetailDefinitionList: React.FC<DetailDefinitionListProps> = ({ entries }) => (
   <DefinitionList>
-    {entries.map(({ term, description, link }) => (
+    {entries?.map(({ term, description, link, alert }) => (
       <DefinitionListItem term={term} key={term}>
+        {/* eslint-disable-next-line no-nested-ternary */}
```
janjaap, line 19: Nested ternaries are very hard to read. Can you rewrite the block below to something more readable?

src/app/pages/MapPage/detail/DetailPanel.tsx
```diff
@@ -143,15 +148,26 @@ const DetailPanel: React.FC<DetailPanelProps> = ({ detailUrl }) => {
   )
 }
 
-const Item: React.FC<{ item: DetailResultItem }> = ({ item }) => {
+const GroupedItems: FunctionComponent<{ item: DetailResultItemGroupedItems }> = ({ item }) => (
+  <>
```
janjaap, line 152: No need for a Fragment here

src/app/pages/MapPage/detail/DetailPanel.tsx
```diff
@@ -143,15 +148,26 @@ const DetailPanel: React.FC<DetailPanelProps> = ({ detailUrl }) => {
...
+    {item.entries.map((groupedItem) => (
+      <>
```
janjaap, line 154: Specify a `key` prop

src/app/pages/MapPage/detail/DetailPanel.tsx
```diff
@@ -161,18 +177,18 @@ const Item: React.FC<{ item: DetailResultItem }> = ({ item }) => {
...
-  // Todo: remove Heading type
-  return component && item.type !== DetailResultItemType.Heading ? (
+  return component ? (
     <div>
       {item.title && (
         <HeadingWrapper>
-          {/*
-          // @ts-ignore */}
-          <DetailHeading styleAs="h4">{item.title}</DetailHeading>
+          <DetailHeading styleAs={subItem ? 'h6' : 'h4'}>{item.title}</DetailHeading>
```
janjaap, line 191: `h6` Is very deeply nested. Does that still result in a semantically correct document?

src/map/components/detail-result/MapDetailResult.tsx
```diff
@@ -1,103 +1,77 @@
...
     >
-      {result.notifications?.map((notification) => {
-        // TODO: This should be refactored so that a notification always has a string value.
-        if (typeof notification.value === 'boolean') {
-          return null
-        }
-
-        return (
+      {result?.data?.notifications?.map((notification) =>
+        notification.value ? (
```
janjaap, line 45: Why not filter `result?.data?.notifications` before mapping? Then this check isn't necessary.

src/app/pages/DataDetailPage/DataDetailPage.tsx
chrisvanmook: will remove 👍

src/app/pages/DataDetailPage/DataDetailPage.tsx
chrisvanmook: True, attribute can be removed

src/app/pages/MapPage/detail/DetailPanel.tsx
chrisvanmook: Thought so too, but then I get this TS error... 
image

I think it does need a Fragment because we iterate and return elements without a parent otherwise

src/app/pages/DataDetailPage/useDataDetail.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      dispatch(
+        showDetail({
+          display: mapDetail._display,
+          geometry,
+        }),
+      )
+      // When a detail doesn't have a geometry, it can only be displayed in VIEWMODE.FULL
+      // Some endpoints only return a geometry when the user is authenticated e.g. authorized to view it
+      if (mapDetail.isAuthorized && !geometry) {
```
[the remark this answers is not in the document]
chrisvanmook: apparently? I basically copy / paste this from a saga so i'd rather leave this in. This whole file can be removed after releasing the new map anyway :)

src/map/services/map-services.config.ts
```diff
@@ -1370,74 +1782,398 @@ const servicesByEndpointType: { [type: string]: ServiceDefinition } = {
...
+              },
+              {
+                term: 'Adres',
+                description: result.verblijfsobjecten
+                  .map(({ _display }: any) => _display)
+                  .join(', '),
+              },
               {
-                value:
-                  'Medewerkers/ketenpartners van Gemeente Amsterdam kunnen inloggen om kadastrale subjecten te bekijken. Om ook zakelijke rechten van natuurlijke personen te bekijken, moet je als medewerker bovendien speciale bevoegdheden hebben.',
```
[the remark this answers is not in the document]
chrisvanmook: They should show up in `PromiseResult`. I've made a ticket in Jira: DI-1204

src/app/pages/MapPage/detail/DetailDefinitionList.tsx
```diff
@@ -1,23 +1,34 @@
...
-import React from 'react'
+import { Alert, CustomHTMLBlock, Link, ShowMoreShowLess } from '@amsterdam/asc-ui'
+import React, { FunctionComponent } from 'react'
+import styled from 'styled-components'
 import { DetailResultItemDefinitionListEntry } from '../../../../map/types/details'
 import DefinitionList, { DefinitionListItem } from '../../../components/DefinitionList'
 
 export interface DetailDefinitionListProps {
-  entries: DetailResultItemDefinitionListEntry[]
+  entries?: DetailResultItemDefinitionListEntry[]
```
jonkoops, line 8: Should we not make this mandatory and prevent empty values from being passed instead? I don't see why anyone would want a list with no entries. If we want to keep this around let's not render the `DefinitionList` at all:

```tsx
const DetailDefinitionList: FunctionComponent<DetailDefinitionListProps> = ({ entries }) => {
  if (!entries) {
    return null
  }

  return (
    <DefinitionList />
  )
}
```

src/app/pages/MapPage/detail/DetailLinkList.tsx
```diff
@@ -29,7 +29,7 @@ const DetailLinkList: React.FC<DetailLinkListProps> = ({ item, ...otherProps })
     })
   }
 
-  return item.links.length ? (
+  return item.links?.length ? (
```
jonkoops, line 32: Should we not assume that links are always present and handle possibly empty values upstream?

src/map/types/details.ts
```diff
@@ -64,51 +70,43 @@ export interface DefaultDetailResultItem {
   title?: string | null
 }
 
-// TODO: Drop 'DetailResultItemDefault' in favor of 'DetailResultItemDefinitionList'
```
jonkoops, line 67: Nice that we can finally drop this one.

src/shared/services/coordinate-reference-system/convertCoordinates.ts
```diff
@@ -0,0 +1,38 @@
+import { rdToWgs84, wgs84ToRd } from './crs-converter'
+
+type Coordinate = [number, number]
```
jonkoops, line 3: Instead of using this type we could also use the `Position` type from the `geojson` package.

src/shared/services/coordinate-reference-system/convertCoordinates.ts
```diff
@@ -0,0 +1,38 @@
...
+
+function convertCoordinates(location: Coordinate, type: 'RD' | 'WGS84') {
+  let wgs84Location
+  let rdLocation
+
+  if (location.length !== 2) {
```
jonkoops, line 9: This check does not reflect the actual type of `Coordinate`, either `Coordinate` needs to be updated to reflect all possible types (e.g. `Position` from the `geojson` package), or this check should be removed.

src/app/pages/MapPage/detail/DetailPanel.tsx
chrisvanmook: `styleAs` will only give the style of a h6 to the component.

src/app/pages/MapPage/detail/DetailDefinitionList.tsx
chrisvanmook: Sure, any preference of how to render it?I think if we do like this, it isn't improving readability much tbh:
```jsx
<DefinitionListItem term={term} key={term}>
          {(() => {
            if (link) {
              return (
                <Link href={link} inList>
                  {description}
                </Link>
              )
            }

            return showMoreShowLess ? (
              <ShowMoreShowLess maxHeight="200px">
                <StyledCustomHTMLBlock body={description} />
              </ShowMoreShowLess>
            ) : (
              description
            )
          })()}
          {alert && <Alert>{alert}</Alert>}
        </DefinitionListItem>
```

src/shared/services/coordinate-reference-system/convertCoordinates.ts
```diff
@@ -0,0 +1,38 @@
+import { rdToWgs84, wgs84ToRd } from './crs-converter'
+
+type Coordinate = [number, number]
+
+function convertCoordinates(location: Coordinate, type: 'RD' | 'WGS84') {
+  let wgs84Location
+  let rdLocation
```
jonkoops, line 7: Instead of defining these values here and giving them a value later on, why not initialize them here directly?

```suggestion
  const wgs84Location = type === 'WGS84' ? location : rdToWgs84({
    x: rdLocation[0],
    y: rdLocation[1],
  })

  const rdLocation = type === 'RD' ? location : wgs84ToRd({
    lat: wgs84Location[0],
    lng: wgs84Location[1],
  })
```

src/app/pages/MapPage/detail/DetailDefinitionList.tsx
chrisvanmook: Or better like this? 
```jsx
const renderShowMoreShowLess = description && description.length > 300
      const renderLink = link && !renderShowMoreShowLess
      const renderDescription = !link && !renderShowMoreShowLess
      return (
        <DefinitionListItem term={term} key={term}>
          {renderShowMoreShowLess && (
            <ShowMoreShowLess maxHeight="200px">
              <StyledCustomHTMLBlock body={description} />
            </ShowMoreShowLess>
          )}
          {renderLink && (
            <Link href={link} inList>
              {description}
            </Link>
          )}
          {renderDescription && description}
          {alert && <Alert>{alert}</Alert>}
        </DefinitionListItem>
```

chrisvanmook: https://github.com/Amsterdam/atlas/pull/3428/commits/7236e6cde5f477a2ea173f091af7f5634753429b

src/shared/services/coordinate-reference-system/convertCoordinates.ts
```diff
@@ -0,0 +1,38 @@
...
+    })
+  } else {
+    wgs84Location = location
+    rdLocation = wgs84ToRd({
+      lat: wgs84Location[0],
+      lng: wgs84Location[1],
+    })
+  }
+
+  const formattedWgs84Location = Object.values(wgs84Location)
```
jonkoops, line 27: `Object.values` has no guarantee that the order remains the same, in this regard it is the same as a for in loop according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in):

> A `for...in` loop iterates over the properties of an object in an arbitrary order (see the delete operator for more on why one cannot depend on the seeming orderliness of iteration, at least in a cross-browser setting).

This means that it cannot be used for this instance where the order of the coordinates is relevant.

src/shared/services/coordinate-reference-system/convertCoordinates.ts
jonkoops, line 7: `wgs84Location` and `rdLocation` are untyped and the value is polymorphic (both an array of numbers or an object with specific properties). A type should be added here.

jonkoops (changes requested):
Going to leave a partial review now so some comments can be addressed.

src/app/components/PromiseTemplate/PromiseResult.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  if (result.status === PromiseStatus.Fulfilled) {
+    return children({ result })
+  }
+
+  if (result.status === PromiseStatus.Pending) {
+    return <StyledLoadingSpinner />
+  }
+
+  if (result.error instanceof AuthError && result.error.code === 401) {
```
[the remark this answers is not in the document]
chrisvanmook: I think that's a good idea, I've created a new ticket DI-1221

src/app/components/PromiseTemplate/PromiseResult.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export interface PageTemplateProps<T> {
+  promise: Promise<T>
+  onRetry?: () => void
+  errorMessage?: string
+  children: (result: PromiseFulfilledResult<T>) => ReactElement | null
+}
+
+const onReload = () => {
+  window.location.reload()
```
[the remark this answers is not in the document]
chrisvanmook: Nice one, thanks!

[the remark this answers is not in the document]
janjaap: Since we're using react router, we can use the `useHistory` hook and call 
```jsx
const history = useHistory()
history.go(0)
```
I think 😬

src/app/pages/MapPage/detail/DetailDefinitionList.tsx
chrisvanmook: Thanks I've updated this

src/shared/services/coordinate-reference-system/convertCoordinates.ts
chrisvanmook: For now I'll leave this in, as we are not sure if we'll actually get a location of type `Position`. It's moved from angular code...

src/shared/services/coordinate-reference-system/convertCoordinates.ts
chrisvanmook: Nice one, was brainlessly copy pasting from old angular junk

src/shared/services/coordinate-reference-system/convertCoordinates.ts
chrisvanmook: Wow didn't knew this. Will make it a string literal instead

src/shared/services/coordinate-reference-system/convertCoordinates.ts
chrisvanmook: It's typed now when applying your suggestion from earlier

src/shared/services/coordinate-reference-system/convertCoordinates.ts
```diff
@@ -0,0 +1,38 @@
+import { rdToWgs84, wgs84ToRd } from './crs-converter'
+
+type Coordinate = [number, number]
+
+function convertCoordinates(location: Coordinate, type: 'RD' | 'WGS84') {
```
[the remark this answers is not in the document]
chrisvanmook: Changed

jonkoops (approved):

PR #3462, opened 2020-10-19
Fixed several bugs for gebieden detail pages

chrisvanmook:


src/app/components/DataSelection/DataSelectionFilters.tsx
```diff
@@ -132,7 +132,7 @@ const DataSelectionFilters: FunctionComponent<Props> = ({
                           }
                         >
                           <span className="c-data-selection-available-filters__item-label">
-                            <span className="qa-option-label">{option.label}</span>
+                            <span className="qa-option-label">{option.label || '(geen)'}</span>
```
jonkoops, line 135: Would also be possible to use the `??` operator here.

```suggestion
                            <span className="qa-option-label">{option.label ?? '(geen)'}</span>
```

src/app/pages/MapPage/detail/DetailDefinitionList.tsx
```diff
@@ -43,7 +37,7 @@ function renderDescription(description: string, link?: LocationDescriptor | null
     )
   }
 
-  if (description.length > 300) {
+  if (description && description.length > 300) {
```
jonkoops, line 40: We can use optional chaining here.

```suggestion
  if (description?.length > 300) {
```

jonkoops (changes requested):

jonkoops (approved):

src/app/pages/MapPage/detail/DetailDefinitionList.tsx
chrisvanmook: Object is possibly `undefined` dus kan niet..

PR #3490, opened 2020-10-22
[DI-1194] Coverage on javascript and typescript

janjaap:
This PR enables coverage generation on both `.js(x)` and `.ts(x)` files. Because all files get coverage now, the thresholds needed to be lowered drastically.


package.json
```diff
@@ -232,12 +233,10 @@
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
       "prettier --write",
-      "eslint",
-      "git add"
```
jonkoops, line 236: Why has the `git add` entry been removed?

.eslintignore
```diff
@@ -2,3 +2,5 @@
 /scripts/
 /test/dyson/
 **/__snapshots__/
+.eslintrc.js
+jest.config.js
```
jonkoops, line 6: Why are we excluding these files exactly? They should also conform the code style.

package.json
```diff
@@ -19,9 +19,10 @@
     "start": "cross-env TS_NODE_PROJECT=tsconfig.webpack.json webpack serve --config webpack.dev.ts",
     "start:prod": "cross-env TS_NODE_PROJECT=tsconfig.webpack.json webpack serve --config webpack.prod.ts",
     "storybook": "start-storybook -p 6007",
-    "test": "concurrently \"npm run test:karma\" \"jest\"",
-    "test:jest": "jest --coverage",
-    "test:jest:watch": "npm run test:jest -- --watch --onlyChanged",
+    "pretest": "rimraf ./coverage ./coverage-legacy",
```
jonkoops, line 22: @chrisvanmook Take note to rebase and disable this on the `remove-angular` branch.

package.json
```diff
@@ -19,9 +19,10 @@
...
+    "test": "concurrently \"npm run test:karma\" \"npm run test:jest\"",
+    "test:jest": "npx jest --coverage",
```
jonkoops, line 24: Why use `npx` here? Jest has already been included as a dev dependency.

jonkoops (changes requested):

tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "./src/**/*"
   ],
   "compilerOptions": {
+    "inlineSourceMap": true,
```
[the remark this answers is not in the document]
janjaap: Doesn't seem required, indeed. I'll remove it.

package.json
janjaap: `lint-staged` Complains about the command, because it already takes care of adding changed files.

package.json
janjaap: No particular reason. Mostly because I always use that on the command line. I'll change it.

.eslintignore
janjaap: The linter complained with the error `Definition for rule 'import/extensions' was not found` when I didn't ignore those files. However, I just removed said line and the linter didn't complain. Removed them.

package.json
jonkoops: Good, then this is resolved.

jonkoops (approved):
LGTM :+1:

PR #3491, opened 2020-10-22
[DI-1109] Bouwdossier linklist fix

janjaap:
This PR fixes an issue where, on bouwdossier pages, a list of (seemingly) duplicate links were shown. The cause of this was that, from the API output, both `nummeraanduiding` results and `verblijfsobject` results were rendered as links to corresponding pages. After making some enquiries at the GOB-team, they advised not to show the `nummeraanduiding` results, because those are subject to change and cannot be referenced as 'address'.

In addition, this PR adds a fixture for the output of the `bouwdossier` endpoint of the `iiif-metadata` API. Note that the fixture data needed to be exported as an explicitly casted type to let typescript deal with string literals (the string 'PUBLIC' could not be correctly typed).


test/withAppContext.js
```diff
@@ -0,0 +1,23 @@
...
+
+import configureStore from '../src/store/store'
+
+const { store, history } = configureStore()
+
+/**
+ * HOC that wraps a component in the necessary providers so that the component can access theming,
+ * the store and can use routing.
+ */
+const withAppContext = (Component) => (
```
jonkoops, line 15: Can we add the correct type information to this method?

src/normalizations/construction-files/getAddresses.ts
```diff
@@ -14,83 +14,79 @@ function formatAddress(
...
+  openbareruimte_id: string
+  huisnummer_letter?: string
+  huisnummer_toevoeging?: string
+  locatie_aanduiding?: string
 }
 
 type AddressResult = {
   id: string
-  type: 'nummeraanduiding' | 'verblijfsobject'
+  type: 'verblijfsobject'
```
jonkoops, line 46: Are we sure this is the only possible type?

src/normalizations/construction-files/getAddresses.ts
```diff
@@ -14,83 +14,79 @@ function formatAddress(
...
-              ),
-          },
-        ],
-        [],
-      ),
-    ],
-    [],
-  )
-}
+const getAddresses = (results: Adressen[]): AddressResult[] =>
```
jonkoops, line 50: The return type can be inferred here.

src/app/components/Gallery/Gallery.tsx
```diff
@@ -69,7 +69,7 @@ type GalleryProps = {
 }
 const MAX_LENGTH = 6
 
-const Gallery: React.FC<GalleryProps> = ({ allFiles, id, access }) => {
+const Gallery: React.FC<GalleryProps> = ({ allFiles, id, access, ...rest }) => {
```
jonkoops, line 72: Conventionally we can `...rest` `...otherProps` in a lot of our code.

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
```diff
@@ -58,7 +77,7 @@ const StyledDefinitionListItem = styled(DefinitionListItem)`
   padding-left: ${themeSpacing(5)}; // Align the terms on the left with the page content
 `
 
-const ConstructionFileDetail: React.FC<ConstructionFileDetailProps> = ({
+const ConstructionFileDetail: FunctionComponent<Bouwdossier> = ({
```
jonkoops, line 80: It's a bit weird we are passing all these props individually instead of just passing one prop in that contains all the data. This is existing code so there is not much we can do about it but it seems like this could be improved in the future.

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
```diff
@@ -13,41 +13,60 @@ export type ConstructionFileImage = {
...
-type ConstructionFile = {
-  barcode: string
-  bestanden: Array<ConstructionFileImage>
-  subdossier_titel: string
-  access: 'RESTRICTED' | 'PUBLIC'
-  document_omschrijving?: string
-  oorspronkelijk_pad?: string
-}
+type Access = 'RESTRICTED' | 'PUBLIC'
+type Status = 'Aanvraag' | 'Behandeling' | null
```
jonkoops, line 17: It might be better to make the `null` overload part of the definition of the field inside the interface, this way this type can be re-used by other interfaces without inheriting the nullability.

jonkoops, line 17: Since these fields are only used in `Bouwdossier` we could name them `BouwdossierAccess` and `BouwdossierStatus` since they are not really re-uasable. Inlining would also be an option, but this is all a stylistic choice.

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
```diff
@@ -13,41 +13,60 @@ export type ConstructionFileImage = {
...
 
-export type ConstructionFileDetailProps = {
+export type Bouwdossier = {
```
jonkoops, line 19: If you feel up for it it would be nice to move all this type information to the `api` directory as per file structure there. This would also mean moving the fetching logic from `ConstuctionFilesPage` in the same directory. Take a look at the existing code there it will give you a good idea of the structure and how these things are put together.

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
```diff
@@ -72,40 +91,41 @@ const ConstructionFileDetail: React.FC<ConstructionFileDetailProps> = ({
...
-            barcode,
-            bestanden: files,
-            subdossier_titel: documentTitle,
-            access: documentAccess,
-            document_omschrijving: description,
-            oorspronkelijk_pad: filePath,
-          }) => (
-            <React.Fragment key={barcode}>
+      {documenten
+        .sort((a, b) => a.subdossier_titel.localeCompare(b.subdossier_titel))
```
jonkoops, line 115: This sorting logic might potentially be expensive to do, should we memoize this value?

jonkoops (changes requested):

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
janjaap: How would I do that?

chrisvanmook: Also would probably make an enum?

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
```diff
@@ -13,41 +13,60 @@ export type ConstructionFileImage = {
...
+type Documenten = {
+  subdossier_titel: string
+  barcode: string
+  bestanden: Bestanden[]
+  oorspronkelijk_pad: string[]
   document_omschrijving?: string
+  access: Access
 }
 
+type Bestanden = {
```
chrisvanmook, line 50: This is a generic problem in our codebase: the use of Dutch and English in code is very random and is driving me crazy 😛  We should make guidelines on how and when we are using Dutch and English (Like `Bouwdossier` vs. `ConstructionFile`). Let's create an issue where we can discuss this, document (ADR?) and have something to refer to. 

In this case for `Bestanden` I think we could use `File` instead, but again, might be something we should discuss first. Also note the type in this case is supposed to be singular, not plural.

src/normalizations/construction-files/getAddresses.test.ts
```diff
@@ -0,0 +1,23 @@
+import bouwdossierFixture from '../../../test/fixtures/iiif-metadata_bouwdossier.json'
+import getAddresses from './getAddresses'
+
+describe('getAddresses', () => {
+  it('returns a sorted list of objects', () => {
```
chrisvanmook, line 5: Think we should return a sorted list of addresses (instead of just objects), right?

test/fixtures/iiif-metadata_bouwdossier.json
```diff
@@ -0,0 +1,90 @@
+{
```
chrisvanmook, line 1: Bit nit-picky here, but isn't this supposed to be a mock, not a fixture?

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
jonkoops: @janjaap remove the `null` here and change `dossier_status?: Status` to `dossier_status?: Status | null` :)

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
jonkoops: I do agree generally, but I would make an exception for the `api` directory as we're trying to map as closely to the terminology of the actual API.

janjaap: I do agree with Jon on this one; we shouldn't try to translate properties from an API endpoint. This will be confusing and make debugging/understanding the code base a lot more difficult.

test/fixtures/iiif-metadata_bouwdossier.json
janjaap: We're mocking the API response with data coming from a fixture.

src/api/iiif-metadata/bouwdossier/getBouwdossier.ts
```diff
@@ -0,0 +1,13 @@
...
+import { fetchWithToken } from '../../../shared/services/api/api'
+import { Bouwdossier } from './types'
+
+/**
+ * Retrieve iiif-metadata about bouwdossiers
+ *
+ * API documentation: https://api.data.amsterdam.nl/iiif-metadata/bouwdossier
+ */
+// eslint-disable-next-line import/prefer-default-export
+export const getBouwdossier = async (id: string) =>
```
jonkoops, line 12: Can we rename this method to `getBouwdossierById`? We're already doing this for some other API methods.

jonkoops, line 12: Just realized since we're not awaiting anything in this method and a promise is returned directly this method does not have to be `async`. Same goes for some of the existing API methods we already have.

src/api/iiif-metadata/bouwdossier/getBouwdossier.ts
```diff
@@ -0,0 +1,13 @@
...
+  fetchWithToken<Bouwdossier>(joinUrl([environment.API_ROOT, 'iiif-metadata', 'bouwdossier', id]))
```
jonkoops, line 13: If you want you could already combine some of the parts of the path that are known. For example.
```suggestion
  fetchWithToken<Bouwdossier>(joinUrl([environment.API_ROOT, 'iiif-metadata/bouwdossier', id]))
```

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
chrisvanmook: Good point, lets only change it to `Bestand` then

src/api/iiif-metadata/bouwdossier/types.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable camelcase */
+export type BouwdossierStatus = 'Aanvraag' | 'Behandeling'
+export type BouwdossierAccess = 'RESTRICTED' | 'PUBLIC'
+
+export type Bestanden = {
```
jonkoops, line 5: Can we use the singular form here for the type name? Same goes for the other types in this file.

src/api/iiif-metadata/bouwdossier/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  openbareruimte_id: string
+  panden: string[]
+  straat: string
+  verblijfsobjecten_label: string[]
+  verblijfsobjecten: string[]
+}
+
+export type Bouwdossier = {
+  access: BouwdossierAccess
+  activiteiten: any[]
```
jonkoops, line 36: Is there a reason why we have `any` here?

src/app/components/ConstructionFileDetail/ConstructionFileDetail.test.tsx
```diff
@@ -14,6 +14,9 @@ jest.mock('../Gallery/Gallery', () => ({ allFiles, ...rest }: { allFiles: any })
   <span {...rest} />
 ))
 
+// convert JSON data to JSON data so that the typescript compiler can parse it (not doing so will show a warning on the use of string literal types)
```
jonkoops, line 17: Instead of doing this why not put the raw data in a `.ts` file so it can infer the types directly?

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
```diff
@@ -1,53 +1,30 @@
 /* eslint-disable camelcase */
 import { Heading, Link, List, ListItem, themeColor, themeSpacing } from '@amsterdam/asc-ui'
-import React from 'react'
+import React, { FunctionComponent, useMemo } from 'react'
 import RouterLink from 'redux-first-router-link'
 import styled from 'styled-components'
-import getAddresses, { Address } from '../../../normalizations/construction-files/getAddresses'
+import getAddresses from '../../../normalizations/construction-files/getAddresses'
 import { toDataDetail } from '../../../store/redux-first-router/actions'
+import { Bouwdossier as BouwdossierType } from '../../../api/iiif-metadata/bouwdossier'
```
jonkoops, line 8: Why create an alias called `BouwdossierType` here? It looks like it doesn't cause any naming conflicts.

src/app/pages/ConstructionFilesPage/ConstructionFilesPage.tsx
```diff
@@ -1,19 +1,20 @@
...
-import ConstructionFileDetail, {
-  ConstructionFileDetailProps as Results,
-} from '../../components/ConstructionFileDetail/ConstructionFileDetail'
+import ConstructionFileDetail from '../../components/ConstructionFileDetail/ConstructionFileDetail'
 import LoadingSpinner from '../../components/LoadingSpinner/LoadingSpinner'
 import useDocumentTitle from '../../utils/useDocumentTitle'
+import usePromise, { PromiseStatus } from '../../utils/usePromise'
+import {
+  getBouwdossier,
+  Bouwdossier as BouwdossierType,
```
jonkoops, line 16: Also here you could use `Bouwdossier` directly without creating an alias.

src/app/pages/ConstructionFilesPage/ConstructionFilesPage.tsx
```diff
@@ -47,24 +48,22 @@ const ConstructionFilesPage: FunctionComponent = () => {
...
       setErrorMessage(ERROR_MESSAGE)
+      return
     }
-    setLoading(false)
-  }
 
-  useEffect(() => {
-    fetchConstructionFiles()
-  }, [id])
+    setResults(bouwdossierResult.value)
```
jonkoops, line 65: The advantage of using `usePromise` is that you already have the value, so there is no need to keep more state around. It also makes it easy the different states.

It might be a good idea to use `PageTemplate` as well, which allows you to delegate these states such as loading, error to a generic component. For a good example of this check out `ActualityPage`.

jonkoops (changes requested):

src/normalizations/construction-files/getAddresses.ts
janjaap: According to the GOB-team, yes. We shouldn't be using nummeraanduidingen.

src/api/iiif-metadata/bouwdossier/getBouwdossier.ts
janjaap: What's the difference?

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
janjaap: Done. Moves types to a separate API folder

src/api/iiif-metadata/bouwdossier/types.ts
janjaap: The type could not be determined. I looked up what the expected type is in the API's repo. Changed it.

src/app/components/ConstructionFileDetail/ConstructionFileDetail.test.tsx
janjaap: Good suggestion. I fixed it by exporting the fixture by casting it to an explicit type.

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
janjaap: It's not clear from the `import` statement that we're importing a type, that's why I renamed it here

src/app/pages/ConstructionFilesPage/ConstructionFilesPage.tsx
janjaap: Let's save that for another PR

src/api/iiif-metadata/bouwdossier/getBouwdossier.ts
jonkoops: Stylistic, but doesn't matter either way.

src/app/components/ConstructionFileDetail/ConstructionFileDetail.tsx
jonkoops: Understandable however we do not do this unless we have conflicting names already in scope. Could you change this?

janjaap: Done

jonkoops (approved):

PR #3492, opened 2020-10-22
Replace existing scripts with new demo deployment scripts

jonkoops:


janjaap (approved):
Noted that this is a temporary solution to replace broken deploy scripts

PR #3499, opened 2020-10-23
Fix URL encoding and clean up login code

jonkoops:


modules/shared/services/http-error-registrar/http-error-registrar.factory.test.js
```diff
@@ -1,227 +0,0 @@
-import { ERROR_TYPES } from '../../../../src/shared/ducks/error/error-message'
```
jonkoops, line 1: I didn't fix up these tests since all this code will be removed anyways.

chrisvanmook (approved):

PR #3500, opened 2020-10-23
Improve deploy-demo script + support for older git versions

chrisvanmook:


scripts/deploy-demo.sh
```diff
@@ -7,15 +7,11 @@ if [[ $SEVER_NUMBER -lt 0 || $SEVER_NUMBER -gt 3 ]]
   exit 1
 fi
 
-SOURCE_BRANCH=$(git branch --show-current)
+SOURCE_BRANCH=$(git rev-parse --abbrev-ref HEAD)
```
jonkoops, line 10: Can we keep this line like the original?

scripts/deploy-demo.sh
```diff
@@ -7,15 +7,11 @@ if [[ $SEVER_NUMBER -lt 0 || $SEVER_NUMBER -gt 3 ]]
...
 TARGET_BRANCH="demo$SEVER_NUMBER"
 
 echo "Pushing $SOURCE_BRANCH to $TARGET_BRANCH."
```
jonkoops, line 13: Perhaps we can get rid of this message now that we only have a single line for the push

jonkoops (changes requested):

jonkoops (approved):

PR #3524, opened 2020-10-28
[DI-1154] Zoek redirect fix

janjaap:
This PR fixes an issue where a redirect wasn't working any more. The cause of this was that a utility function expected an API request to return with a specific prop. The API endpoint was changed and the new output didn't have the expected prop.

In order to fix it, the proposed changes:
- add fixtures and an API entrypoint function to the API `v1/bag/nummeraanduiding` folder.
  Note the folder structure; the Amsterdam API contains more than one `nummeraanduiding` endpoint of which we use two: `/bag/v1.1/nummeraanduiding` and `/v1/bag/nummeraanduiding`. The former uses an unsupported naming structure and will be phased out at some point. The latter uses the official (yet not officially released and apparently still in beta 🤯 ) structure.
- add `nummeraanduiding` types for the `/v1/bag/nummeraanduiding` endpoint
- add sufficient tests


src/app/redirects.js
```diff
@@ -192,7 +192,7 @@ export default async function resolveRedirects() {
       })
 
       // Tries to prevent cancelling the network request to Matomo, arbitrary number that allows Matomo some time to load
-      window.setTimeout(() => window.location.replace(redirectTo), 1000)
+      window.setTimeout(() => window.location.replace(redirectTo), 0)
```
alimpens, line 195: Why do you use a timeout of zero seconds here? Doesn't that break this hack?

src/app/redirects.test.js
```diff
@@ -26,12 +26,12 @@ const expectWithNewRoute = async (route, expectation, cb) => {
   try {
     await resolveRedirects()
     cb(() => expect(window.location.replace))
-  } catch (e) {
-    // console.log(e) hide unexpected async errors from Jest
+  } catch {
+    // no-op
```
alimpens, line 30: This is a bit opaque. Maybe define why catch shouldn't do anything?

src/app/redirects.js
janjaap: The whole point of the `setTimeout` call is to delay the `window.location.replace` call. There is no need to delay it with 1000 milliseconds; the code should be executed when all other code has been executed and then a delay of 0 ms is enough since the execution is moved from the execution stack to the event loop.

src/app/redirects.test.js
janjaap: This was existing code. I didn't look into why a `try/catch` block is necessary here.

alimpens: I think the empty catch is to `hide unexpected async errors from Jest`, right? That's a bit more descriptive than `no-op`

janjaap: It's pretty clear that an empty `catch` does that; the comment doesn't add anything we don't already know.

alimpens: I disagree, I think the original comment makes it obvious that we're intentionally not showing something. `no-op` just means nothing is happening there.

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+type ReceiveFields = {
+  _fields: string
+}
+
+/**
+ * Retrieve nummeraanduiding
+ *
+ * API documentation: https://api.data.amsterdam.nl/v1/bag/nummeraanduiding
+ *
+ * @param {String} queryParams - Full URL or search string
```
jonkoops, line 16: There is no need to annotate the type here as we already get this information from TypeScript.

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+    return null
+  }
+
+  if (receiveFields) {
+    searchParams.append('_fields', receiveFields)
+  }
+
+  const url = joinUrl([environment.API_ROOT, 'v1/bag/nummeraanduiding'])
+
+  return fetchWithoutToken(`${url}?${searchParams.toString()}`)
```
jonkoops, line 37: If you do not want to do string concatenation I believe you could also use the `URL` constructor here. 
```suggestion
  const url = new URL(joinUrl([environment.API_ROOT, 'v1/bag/nummeraanduiding']))
  
  url.search  = searchParams.toString()

  return fetchWithoutToken(url.toString())
```

src/app/utils/getVerblijfsobjectIdFromAddressQuery.ts
```diff
@@ -0,0 +1,23 @@
+import { getNummeraanduidingByAddress } from '../../api/v1/bag/nummeraanduiding'
+
+/**
+ * Get a verblijfsobject id from a URL or search string
+ *
+ * @param {String} locationSearch - Full URL or search string
```
jonkoops, line 6: No need to add the type here as well, this will be taken care of by TypeScript.

src/app/utils/getVerblijfsobjectIdFromAddressQuery.ts
```diff
@@ -0,0 +1,23 @@
...
+  const result = await getNummeraanduidingByAddress(locationSearch)
+
+  if (!result) return ''
+
+  const {
+    page: { totalElements },
+    _embedded: data,
+  } = result
+
+  if (totalElements !== 1) return ''
```
jonkoops, line 18: Instead of checking the `totalElements` it would be more safe to check for the value of `data.nummeraanduiding[0]` directly. Unfortunately TypeScript will always return a non-undefined type from array accessors but that of course does not mean the value returned cannot be undefined.

jonkoops (changes requested):

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+// eslint-disable-next-line import/prefer-default-export
+export const getNummeraanduidingByAddress = (
+  queryParams: string,
+  receiveFields?: string,
+): Promise<Nummeraanduiding> | null => {
+  const paramsString = queryParams.substr(queryParams.indexOf('?'))
+  const searchParams = new URLSearchParams(queryStringParser(paramsString))
+
+  if (!searchParams.get('postcode') || !searchParams.get('huisnummer')) {
```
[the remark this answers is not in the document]
janjaap: Could do that, but then it will pass for `&postcode=&huisnummer=` in the URL, because `.has()` will return true for an empty string.

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
janjaap: I know, but I added the docblock so that I could mention that the function can expect either a full url or just the search string. That is not something that can be deducted from the type

src/app/utils/getVerblijfsobjectIdFromAddressQuery.ts
janjaap: Again true, but I wanted to indicate that the function can receive both a full url as well as a search string

src/app/utils/getVerblijfsobjectIdFromAddressQuery.ts
janjaap: Not entirely true, because the function needs to return a value only when the total amount equals 1. I can check for the value of the first element in the result, but I still need to ensure that there is only one element.

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
janjaap: I don't see a benefit with the proposed change.

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @param {String} queryParams - Full URL or search string
+ * @param {String} receiveFields - comma separated (no spaces) values indicating which fields should be returned from the API request
+ */
+// eslint-disable-next-line import/prefer-default-export
+export const getNummeraanduidingByAddress = (
+  queryParams: string,
+  receiveFields?: string,
+): Promise<Nummeraanduiding> | null => {
+  const paramsString = queryParams.substr(queryParams.indexOf('?'))
```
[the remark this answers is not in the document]
janjaap: `queryParams` Can be either a full url or just a search string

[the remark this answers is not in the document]
jonkoops: Then this point is indeed moot.

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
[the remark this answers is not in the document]
jonkoops: Ah, yes. That makes sense :+1:

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
jonkoops: True, but you can write the same docs without the type as follows:

```
@param queryParams - Full URL or search string
```

This will work also. I actually thought we had a linting rule for this but it might only be on other repos.

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
jonkoops: The benefit is that you are leaving *all* the parsing and formatting logic for URLs to the built-in classes that are provided by the browser.

src/app/utils/getVerblijfsobjectIdFromAddressQuery.ts
jonkoops: Same as the previous one, you can omit the type and keep the rest :smile:

src/app/utils/getVerblijfsobjectIdFromAddressQuery.ts
jonkoops: Ah I was not aware of this logic, then this code makes sense :+1:

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
```diff
@@ -0,0 +1,33 @@
...
+// eslint-disable-next-line import/prefer-default-export
+export const getNummeraanduidingByAddress = (
+  queryParams: string,
+  receiveFields?: string,
+): Promise<Nummeraanduiding | null> => {
+  const paramsString = queryParams.substr(queryParams.indexOf('?'))
+  const searchParams = new URLSearchParams(paramsString)
+
+  if (!searchParams.get('postcode') || !searchParams.get('huisnummer')) {
+    return null
```
jonkoops, line 23: Looks like the build is failing since this is not a `Promise<null>`. You can either mark this method as `async` or return `Promise.resolve(null)` here.

jonkoops (changes requested):

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
janjaap: In your example, search param parsing happens twice, because of `searchParams.toString()` and setting the result of that to `url.search`. I'm not inclined to apply the proposed change here.

jonkoops (approved):

src/api/v1/bag/nummeraanduiding/getNummeraanduidingByAddress.ts
jonkoops: It's not that important, either way is fine. Parsing should not occur twice per-se, that all depends on the internal API of `URL`. Besides that, it's not like we're calling this method on any budget so it doesn't really matter if it was parsed multiple times.

PR #3536, opened 2020-10-29
(Chore) isAuthenticated

janjaap:
This PR adds the `isAuthenticated` function to the `auth` module. This allows us to do the following:
```javascript
import { isAuthenticated, getAccessToken } from '/shared/services/auth/auth'

const fetchData = async (requestURL: string, headers: any, options: any) => {
  const authHeaders = isAuthenticated() ? { Authorization: `Bearer ${getAccessToken()}` } : {}
  const requestHeaders = {
    'Content-Type': 'application/json',
    Accept: 'application/json',
    ...headers,
    ...authHeaders,
  }

  // handle fetch
}
```

It also allows us to shield off specific parts of the application by just checking the return value of `isAuthenticated()`. Another benefit is that, when proxying fetch requests in tests, it will allow us to provide test cases with specific data based on the return value of `isAuthenticated()` since some endpoints return different outputs based on the authentication level of a user.

### DIFF
--- a/src/shared/services/auth/auth.test.js
+++ b/src/shared/services/auth/auth.test.js
@@ -6,7 +6,16 @@ import joinUrl from '../../../app/utils/joinUrl'
 import environment from '../../../environment'
 import queryStringParser from '../query-string-parser/query-string-parser'
 import stateTokenGenerator from '../state-token-generator/state-token-generator'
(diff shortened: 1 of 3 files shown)


chrisvanmook:
I think later we should really revise the whole `auth.ts` + `api.ts` files... Perhaps a hook in our case, where we can handle authentication as well

chrisvanmook (approved):

PR #3552, opened 2020-11-02
(Chore) Fetch proxy

janjaap:
This PR adds a function to the API service that will automatically append authentication headers to requests. This will (in an upcoming PR) replace all `fetchWithToken` and `fetchWithoutToken` calls.


src/shared/services/api/api.test.js
```diff
@@ -1,11 +1,9 @@
-import { createUrlWithToken, fetchWithToken } from './api'
+import { createUrlWithToken, fetchWithToken, fetchProxy } from './api'
```
jonkoops, line 1: Can we turn this into a TypeScript file while we are at it?

src/shared/services/api/api.ts
```diff
@@ -75,3 +80,27 @@ export const createUrlWithToken = (url: string, token: string) => {
...
+    method: 'GET',
+  }
+
+  const searchParams = new URLSearchParams(params)
+
+  const fullUrl = new URL(url)
+  fullUrl.search = searchParams.toString()
+
+  return fetch(fullUrl.href, options)
+    .then((response) => handleErrors(response, reloadOnUnauthorized))
```
jonkoops, line 104: As far as I can tell the `reloadOnUnauthorized` flag is never used anywhere, let's remove all instances of it from this file (including the other methods).

src/shared/services/api/api.ts
```diff
@@ -75,3 +80,27 @@ export const createUrlWithToken = (url: string, token: string) => {
 
   return parsedUrl.toString()
 }
+
+export const fetchProxy = <T = any>(
```
jonkoops, line 84: Why the name `fetchProxy`? What is this method proxying?

src/shared/services/api/api.ts
```diff
@@ -1,8 +1,13 @@
...
+import { logout, getAuthHeaders } from '../auth/auth'
 import getState from '../redux/get-state'
 import SHARED_CONFIG from '../shared-config/shared-config'
 import { AuthError, NotFoundError } from './errors'
 
+type FetchProxy = {
+  params?: UrlParams
+  headers?: Headers
+  reloadOnUnauthorized?: boolean
+}
```
jonkoops, line 10: I'd like to see this be as close as possible to the existing Fetch API, let's extend the existing `RequestInit` interface here and add the functionality we want to use to make our lives easier. For example:

```ts
interface FetchOptions extends RequestInit {
  searchParams?: UrlParams
}
```

Then inside of the method where we use this interface we can do:

```ts
function fetch(url: string, options?: FetchOptions) {
  const { searchParams, ...fetchOptions } = options ?? {}
  const fullUrl = new URL(url)

  fullUrl.search = new URLSearchParams(searchParams).toString()

  return fetch(fullUrl.toString(),  fetchOptions)
}
```

src/shared/services/api/api.ts
```diff
@@ -75,3 +80,27 @@ export const createUrlWithToken = (url: string, token: string) => {
...
+  { params, headers, reloadOnUnauthorized = false }: FetchProxy = {},
+): Promise<T> => {
+  const requestHeaders = new Headers(headers)
+
+  const options: RequestInit = {
+    headers: {
+      ...requestHeaders,
+      ...getAuthHeaders(),
+    },
+    method: 'GET',
```
jonkoops, line 95: `GET` is the default, no need to specify it.

jonkoops (changes requested):

src/shared/services/api/api.ts
janjaap: Well, `fetch` was already taken and I needed a name. In essence, this function proxies the global `fetch` function, but I'm open to changing it if you have a better suggestion.

src/shared/services/api/api.ts
janjaap: There is one function that sets `reloadOnUnauthorized` to true; `geosearch` in `/shared/services/search`

src/shared/services/api/api.ts
jonkoops: I actually do not have any better suggestions, just wondering as to the motivation. Let's leave it like this, we can always rename it in the future.

src/shared/services/api/api.ts
jonkoops: Let's remove the one from `fetchProxy` only then, I think that kind of stuff should be handled upstream.

src/shared/services/api/api.ts
janjaap: I'm also not too happy with the name, but it was the best I could think of.

src/shared/services/api/api.ts
```diff
@@ -75,3 +78,25 @@ export const createUrlWithToken = (url: string, token: string) => {
...
 }
+
+export const fetchProxy = <T = any>(url: string, init: FetchOptions = {}): Promise<T> => {
+  const { headers, searchParams, ...otherOptions } = init
+  const requestHeaders = new Headers(headers)
+
+  const options: RequestInit = {
+    ...otherOptions,
+    headers: {
+      ...requestHeaders,
```
jonkoops, line 89: Are you sure we can spread this `Headers` object like this? Does it implement `Iterable`?

src/shared/services/api/api.ts
```diff
@@ -11,7 +14,7 @@ const getAccessToken = () => getState()?.user?.accessToken
 export const fetchWithoutToken = <T = any>(uri: string): Promise<T> =>
   fetch(uri).then((response) => response.json())
 
-const handleErrors = (response: Response, reloadOnUnauthorized: boolean) => {
+const handleErrors = (response: Response, reloadOnUnauthorized: boolean = false) => {
```
jonkoops, line 17: When providing a default value it is not needed to specify the type as it can be inferred.

```suggestion
const handleErrors = (response: Response, reloadOnUnauthorized = false) => {
```

jonkoops (approved):
Looks good to me!

src/shared/services/api/api.ts
janjaap: Good point. Not possible indeed. Applied a different way of merging the headers.

jonkoops (approved):

PR #3565, opened 2020-11-03
Format fields on dataset detail page as markdown

jonkoops:
Some fields expect to be formatted as markdown, this PR makes sure that they do. Impact on bundle size is minimal as `marked` gets loaded into the `DatasetDetailPage` chunk.


package.json
```diff
@@ -61,6 +61,7 @@
     "lodash.debounce": "^4.0.8",
     "lodash.defer": "^4.1.0",
     "lodash.isequal": "^4.5.0",
+    "marked": "^1.2.2",
```
janjaap, line 64: Gzipped size is 11 KB. Do we really need this package? Looking at the schema, the fields used are typed as `rst`: https://github.com/Amsterdam/dcatd/blob/0920e894ab1249e6d8772198da8f3e552fba14d3/dcat_schema.yml#L13. The API output shows me HTML

jonkoops: As discussed we will resolve this by making this API return only HTML in the future.

janjaap (approved):

PR #3582, opened 2020-11-04
[DI-1244] Test serviceworker proxy

janjaap:
This PR provides a way to proxy requests to API endpoints by providing data fixtures that are served by a serviceworker process that captures requests when running tests. (For more information on the approach, read https://kentcdodds.com/blog/stop-mocking-fetch). It comes down to not checking with what `fetch` has returned with, because in nine out of ten cases, this is not important. More important is how the component under test handles the data (or lack of data) and deals with error responses.

This has a couple of advantages:
- API responses are centralized
- Responses always return the same data
- Response fixtures can be typed
- Tests can be simplified
- Components now all use the same (correctly typed) data when rendered in test cases

In this PR, the `getPanoramaThumbnail` tests contain examples on how the new approach can be used when different responses need to be simulated. The `auto-suggest` service contains an example on how centralized API fixtures are dealt with.


jonkoops (approved):

janjaap:
Closing; replaced by #3856

PR #3599, opened 2020-11-09
(Chore) Docker cleanup

janjaap:
The changes in this PR make it easier to set up the application locally, with Docker. 
It:
- reorders commands in the Dockerfile for faster building; putting files that are less likely to change higher up in the command stack
- expands the `.dockerignore` list so that Docker won't include unnecessary files in the build context
-  adds a `docker-compose.yml` file as well as a `Makefile`
- updates the Nginx configuration so that, when the apps runs in http, assets from `fast.fonts.net` will be downloadable
- allows both `debugger` and `console` statements to be left in the production bundle, when the environment variable `DEBUG` has been set with the value 'true'


.dockerignore
```diff
@@ -1,8 +1,13 @@
+.dependabot
```
jonkoops, line 1: Instead of keeping this file up to date like this perhaps we can ignore all files and include only the ones we need. See: https://kevinpollet.dev/posts/how-to-use-your-dockerignore-as-a-whitelist/

Dockerfile
```diff
@@ -1,55 +1,56 @@
...
 WORKDIR /app
 
-COPY package.json package-lock.json .env.* /app/
-COPY sitemap-generator /app/sitemap-generator
-COPY public /app/public
+COPY sitemap-generator ./sitemap-generator
 
+COPY public ./public
+
+COPY package.json package-lock.json /app/
```
jonkoops, line 13: Since the workdir is already the `/app` directory can we just copy these directly?

```suggestion
COPY package.json package-lock.json .
```

Dockerfile
```diff
@@ -1,55 +1,56 @@
...
-    #    npm config set registry https://nexus.data.amsterdam.nl/repository/npm-group/ && \
-    PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true \
-    npm --production=false \
-    --unsafe-perm \
-    --verbose \
-    ci
+    git config --global url."https://github.com/".insteadOf git@github.com:
+
+RUN npm ci && \
+    npm cache clean --force
```
jonkoops, line 19: Why do we want to run a cache clean here?

Dockerfile
```diff
@@ -1,55 +1,56 @@
...
-# Build dependencies
-COPY src /app/src
 COPY modules /app/modules
+
 COPY .babelrc \
-    .eslintrc.js \
-    .eslintignore \
-    .prettierrc \
-    .prettierignore \
+    .browserslistrc \
```
jonkoops, line 26: I don't believe we are actually using the `browserslistrc` file as we're passing this information in the Webpack config manually but I could be wrong.

Dockerfile
```diff
@@ -1,55 +1,56 @@
...
     index.ejs \
     webpack.* \
-    tsconfig.json \
-    tsconfig.webpack.json \
+    tsconfig.* \
     favicon.png \
     /app/
```
jonkoops, line 31: Same here, the workdir is still `/app` so I think we can just copy the files to the current directory.

Dockerfile
```diff
@@ -1,55 +1,56 @@
...
 
+COPY src /app/src
+COPY test /app/test
```
jonkoops, line 34: Same here, the workdir is still `/app` so I think we can just copy the files to the current directory.

Dockerfile
```diff
@@ -1,55 +1,56 @@
...
+
 RUN npm run build
-RUN echo "build= `date`" > /app/dist/version.txt
 
-# Test dependencies
-COPY jest.config.js \
-    /app/
-COPY test /app/test
+ARG BUILD_NUMBER=0
+RUN echo "build ${BUILD_NUMBER} - `date`" > ./dist/version.txt
```
jonkoops, line 42: `BUILD_NUMBER` gets passed in for sure?

Dockerfile
```diff
@@ -1,55 +1,56 @@
...
 
-# Web server image
+################################
+# Deploy
```
jonkoops, line 45: This is not really a deployment.

nginx.conf
```diff
@@ -30,10 +30,10 @@ http {
     add_header Content-Security-Policy
       "default-src 'self';
       connect-src 'self' https://*.data.amsterdam.nl;
-      font-src 'self' 'unsafe-inline' 'unsafe-eval' https://static.amsterdam.nl https://fast.fonts.net;
+      font-src 'self' 'unsafe-inline' 'unsafe-eval' https://static.amsterdam.nl fast.fonts.net;
```
jonkoops, line 33: Should we also remove the protocol of other entries in this file? Are we sure we are still only allowing HTTPS as a protocol by removing it?

webpack.prod.ts
```diff
@@ -39,7 +40,13 @@ export default [
       minimizer: [
         new TerserPlugin({
           terserOptions: {
+            compress: {
+              drop_console: !debugMode,
+              drop_debugger: !debugMode,
+            },
             sourceMap: true,
+            mangle: !debugMode,
+            keep_fnames: debugMode,
```
jonkoops, line 49: Should these two flags not be handled by the sourcemapping in the browser?

Makefile
```diff
@@ -0,0 +1,51 @@
+# show help by default
```
jonkoops, line 1: I personally never use makefiles, what is the added advantage of this vs. just using the Docker Compose CLI tools?

jonkoops (changes requested):

Dockerfile
janjaap: Because we need to keep the intermediate containers as small as possible and therefore we need to remove anything that we don't need in production.

Dockerfile
janjaap: No, doesn't look like it: https://data.amsterdam.nl/version.txt.
I'll remove it.

nginx.conf
janjaap: I removed it, because the local Docker setup runs in http instead of https and leaving this in will throw warnings. We might be able to circumvent this by settings specific flags in Chrome. I'll have a look.

Makefile
janjaap: It acts as a proxy so that you don't have to remember any Docker-related commands. Also, when the setup gets more complex (for whatever reason), we can move complexity into the Makefile.
A Makefile is also a way to extract logic and complexity when running in, for instance, Jenkins; it's easier to issue commands from there instead of having to resort to Groovy or DSL.

Dockerfile
janjaap: That image is deployed, not the one above.

Dockerfile
janjaap: The file is read by Webpack; removing it will prevent the application from building

nginx.conf
janjaap: Because the local environment runs on http, requests will go to `//fast.fonts.net` which equals to `http://fast.fonts.net`. Removing the schema allows requests to either http or https.

jonkoops (approved):

chrisvanmook (approved):

PR #3605, opened 2020-11-10
[DI-985] Implement new detail panel structure

jonkoops:
DI-985


package.json
```diff
@@ -27,8 +27,8 @@
   },
   "dependencies": {
     "@amsterdam/arm-cluster": "^0.4.0",
-    "@amsterdam/arm-core": "^0.4.0",
-    "@amsterdam/arm-draw": "^0.4.0",
+    "@amsterdam/arm-core": "file:.yalc/@amsterdam/arm-core",
```
jonkoops, line 30: This will be replaced once we run a new release of ARM.

src/app/pages/MapPage/legend/LegendPanel.tsx
```diff
@@ -1,54 +1,55 @@
...
-    left: 0;
-
-    & > .scroll-wrapper {
-      max-height: 100%;
-      height: 100%;
-    }
+  & > .scroll-wrapper {
+    height: initial;
+    max-height: initial;
+    overflow: visible;
```
janjaap, line 25: This will show multiple scrollbars; are you sure this rule should be set?

src/app/components/PanoramaViewer/PanoramaViewerControls.tsx
```diff
@@ -17,29 +17,29 @@ type Props = {
...
-        <>
-          <ControlButton
+        <Control>
+          <Button
             type="button"
             variant="blank"
             title="Panorama sluiten"
             size={44}
             iconSize={25}
+            // @ts-ignore
```
janjaap, line 42: Why is it necessary to suppress errors here? If we cannot do without error suppression, can you define on this line which errors are suppressed?

src/app/pages/MapPage/detail/DetailPanel.tsx
```diff
@@ -144,17 +144,19 @@ const DetailPanel: FunctionComponent<DetailPanelProps> = ({ detailUrl }) => {
...
-          <RenderDetails details={value} />
-        </MapPanelContent>
-      )}
-    </PromiseResult>
+    <LargeDrawerPanel {...otherProps}>
+      <PromiseResult promise={promise}>
+        {({ value }) => (
+          <>
+            <DrawerPanelHeader onClose={() => setDetailUrl(null)}>
+              <h2>{value?.data.subTitle ?? 'Detailweergave'}</h2>
```
janjaap, line 152: Is an `h2` semantically correct in this context? I don't see that tag in the original code.

PR #3611, opened 2020-11-11
[DI-1339] Floating point values fix

janjaap:
This PR fixes an issue in meetbout meting tables where floating point values should be shown, but significant numbers are truncated.

### DIFF
--- a/src/api/meetbouten/meting/types.ts
+++ b/src/api/meetbouten/meting/types.ts
@@ -1,0 +1,18 @@
+export interface Root {
+  count: number
+  results: Meting[]
+}
+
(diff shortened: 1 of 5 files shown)


src/api/meetbouten/meting/types.ts
```diff
@@ -0,0 +1,18 @@
+export interface Root {
+  count: number
+  results: Meting[]
+}
+
+// fields zakking, hoogte_nap, zakkingssnelheid and zakking_cumulatief are actually floating point values, but the API returns them as string values 🙄
```
jonkoops, line 6: Facepalm

jonkoops (approved):

PR #3612, opened 2020-11-11
Updated ASC to 0.26.0

alimpens:


src/app/models/notification.ts
```diff
@@ -1,6 +1,6 @@
 export enum NotificationLevel {
```
jonkoops, line 1: Can we get rid of this enum entirely and just use the values to the `Alert` directly? It looks like this enum is now used to express two different concerns.

jonkoops (changes requested):

src/app/models/notification.ts
jonkoops: Let's take a look at this some other time

jonkoops (approved):

PR #3632, opened 2020-11-17
Make promises passed to usePromise lazy loaded

jonkoops:


src/app/utils/usePromise.ts
```diff
@@ -25,26 +25,29 @@ export interface PromiseRejectedResult {
   error: any
 }
 
+export type PromiseFactoryFn<T> = () => Promise<T>
+
 /**
- * Takes a promise and returns it's resolved or rejected value. While the promise is in flight an intermediate value is returned which can be used for loading states.
- *
- * When passing in a promise make sure that the instance is cached using `useMemo`, for example:
+ * Takes a function that creates a promise and returns it's resolved or rejected value together with the status of the promise.
```
janjaap, line 31: its instead of it's

src/app/utils/usePromise.ts
```diff
@@ -25,26 +25,29 @@ export interface PromiseRejectedResult {
...
+ * const result = usePromise(() => fetchUser(userId), [userId])
  * ```
  *
- * @param promise The promise of which the value will be retrieved.
+ * @param factory The function that creates the promise that will be used.
+ * @param deps The dependencies of the factory function.
  */
-export default function usePromise<T = any>(promise: Promise<T>) {
+export default function usePromise<T = any>(factory: PromiseFactoryFn<T>, deps?: DependencyList) {
+  const promise = useMemo(factory, deps)
```
janjaap, line 42: Since `factory` is always a function, shouldn't it be wrapped in `useCallback` instead of `useMemo`?

src/app/utils/usePromise.test.ts
```diff
@@ -82,6 +86,37 @@ describe('usePromise', () => {
...
+    rerender()
+
+    expect(result.current).toEqual({
+      status: PromiseStatus.Fulfilled,
+      value: 0,
+    })
+
+    retryCount = 1
+    rerender()
+    await waitForNextUpdate()
```
janjaap, line 113: Is there a difference between 
```
rerender()
await waitForNextUpdate()
```

and 
```
await waitForNextUpdate()
rerender()
```
?
If not, maybe use the same order to avoid confusion.

src/app/utils/usePromise.ts
jonkoops: In this case we are not memoizing the `factory` function, but the promise that is returned from it. So in this case it seems valid to me to use `useMemo`.

src/app/utils/usePromise.ts
jonkoops: Good catch, changed.

src/app/utils/usePromise.test.ts
jonkoops: The order doesn't really matter much, I've changed this to be the same.

janjaap (approved):

PR #3634, opened 2020-11-17
[DI-1333] service worker

janjaap:
This PR implements caching strategies for assets loaded in the `atlas` application. For the service worker to work correctly, this PR:
- lets Webpack inject the necessary chunk references in the public HTML file
- loads a service worker only when on production
- transpiles chunks without hashes in their file names; the service worker already takes care of versioning
- removes cache groups configuration from Webpack production configuration; this allows Webpack to determine for itself which chunks need to be generated and, subsequently, which can be added to or removed from the cache when a page reload occurs
- defines specific caching strategies for assets that are retrieved from all subdomains that the `atlas` application queries. Each subdomain has its own caching lifetime, based on the nature of the asset that is served from a subdomain (images can be cached for a longer period than JSON responses should be, for instance)

The following scenarios have been tested:
1. No service worker => active service worker
   - A visitor has a version of the site open that doesn't have the service worker applied
   - A release with service worker is deployed
   - The visitor keeps navigating the site; the service worker is picked up seamlessly by the browser

2. Active service worker => new release
   - A visitor has a version of the site open that has the service worker applied
   - A new release is deployed
   - The visitor keeps navigating the site; the changes in the new release are picked up by the service worker and applied correctly

3. Active service worker => no service worker
   - A visitor has a version of the site open that has the service worker applied
   - A new release is deployed where the service worker configuration is removed
   - The visitor keeps navigating the site; the removed service worker doesn't lead to chunks that cannot be found


chrisvanmook (approved):

src/store/store.js
```diff
@@ -7,7 +7,7 @@ import createSagaMiddleware from 'redux-saga'
 import routes from '../app/routes'
 import rootReducer from '../reducers/root'
 import rootSaga from '../root-saga'
-import '../shared/ducks/error/error-message'
+// import '../shared/ducks/error/error-message'
```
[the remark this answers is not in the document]
janjaap: Ah, yes. I came across those and I was wondering what they were doing there. After looking into it, I noticed that those modules only have default or named exports and don't execute any code, but I wasn't entirely sure. Do you know, perhaps?

webpack.common.ts
```diff
@@ -274,8 +274,8 @@ export function createConfig(additionalOptions: CreateConfigOptions): Configurat
         },
       }),
       new MiniCssExtractPlugin({
-        filename: isProd ? '[name].[contenthash].css' : '[name].css',
-        chunkFilename: isProd ? '[name].[contenthash].css' : '[name].css',
+        filename: '[name].css',
+        chunkFilename: '[name].css',
```
jonkoops, line 278: Would this not cache the CSS file forever? Since the name will always be the same, even if the contents have changed.

webpack.dev.ts
```diff
@@ -14,6 +14,7 @@ export default [createConfig({ legacy: false, singleBuild: true, mode: 'developm
         disableHostCheck: true,
         contentBase: distPath,
         compress: true,
+        watchContentBase: true,
```
jonkoops, line 17: Why do we need to watch the build result in `dist`?

src/store/store.js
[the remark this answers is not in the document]
jonkoops: Yeah looks weird, especially the `queryParameters` which seems to register new paramaters on onother module.

webpack.prod.ts
```diff
@@ -2,131 +2,143 @@ import path from 'path'
 import TerserPlugin from 'terser-webpack-plugin'
 import { merge } from 'webpack-merge'
 import { createConfig, srcPath } from './webpack.common'
+import { GenerateSW } from 'workbox-webpack-plugin'
 
-const CHUNKS = {
```
jonkoops, line 6: Good to see this gone.

webpack.prod.ts
```diff
@@ -2,131 +2,143 @@ import path from 'path'
...
-export default [
-  createConfig({ legacy: false, mode: 'production' }),
-  createConfig({ legacy: true, mode: 'production' }),
-].map((config) =>
-  merge(config, {
-    bail: true,
-    resolve: {
-      alias: {
-        [path.resolve(srcPath, 'environment.ts')]: path.resolve(srcPath, 'environment.prod.ts'),
+export default [createConfig({ legacy: false, mode: 'production', singleBuild: true })].map(
```
jonkoops, line 9: Are we only doing non-legacy builds now? Perhaps we should first pick up the story to remove legacy browser support.

webpack.prod.ts
```diff
@@ -2,131 +2,143 @@ import path from 'path'
...
+  (config) =>
+    merge(config, {
+      bail: true,
+      resolve: {
+        alias: {
+          [path.resolve(srcPath, 'environment.ts')]: path.resolve(srcPath, 'environment.prod.ts'),
+        },
+      },
+      output: {
+        filename: '[name].js',
```
jonkoops, line 19: Why are we only using `name` here? Now our assets are no longer fingerprinted and this might cause the wrong code to load.

webpack.prod.ts
```diff
@@ -2,131 +2,143 @@ import path from 'path'
...
-            compress: {
-              drop_console: !debugMode,
-              drop_debugger: !debugMode,
+      devtool: 'source-map',
+      optimization: {
+        minimizer: [
+          new TerserPlugin({
+            terserOptions: {
+              compress: {
+                drop_debugger: !debugMode,
```
jonkoops, line 28: I think we have a linting rule to prevent debugger statements from showing up. This could be removed IMHO.

webpack.prod.ts
```diff
@@ -2,131 +2,143 @@ import path from 'path'
...
-            chunks: 'async',
-            enforce: true,
-          },
-          main: {
-            chunks: 'all',
-            minChunks: 2,
-            reuseExistingChunk: true,
+          }),
+        ],
+        nodeEnv: 'production',
```
jonkoops, line 34: This will default to the `mode` argument, which is set to `production`, this line can be removed.

webpack.prod.ts
```diff
@@ -2,131 +2,143 @@ import path from 'path'
...
+        namedChunks: true,
+        namedModules: true,
```
jonkoops, line 36: These options have been removed in future versions of WebPack, let's not introduce them here. See: https://webpack.js.org/blog/2020-10-10-webpack-5-release/#changes-to-the-structure

webpack.prod.ts
```diff
@@ -2,131 +2,143 @@ import path from 'path'
...
+        moduleIds: 'named',
+        chunkIds: 'named',
```
jonkoops, line 38: According to the WebPack documentation:
> The `deterministic` option is useful for long term caching, but still results in smaller bundles compared to `hashed`
- https://webpack.js.org/configuration/optimization/#optimizationmoduleids

Since this is the production build should we not be using this option instead?

webpack.prod.ts
```diff
@@ -2,131 +2,143 @@ import path from 'path'
...
+        runtimeChunk: false,
+        splitChunks: {
```
jonkoops, line 40: Would the default config not be enough here? See: https://webpack.js.org/plugins/split-chunks-plugin/#optimizationsplitchunks

webpack.prod.ts
```diff
@@ -2,131 +2,143 @@ import path from 'path'
...
+          cacheGroups: {
+            default: false,
           },
         },
       },
-    },
-  }),
+      plugins: [
+        new GenerateSW({
+          mode: debugMode ? 'development' : 'production',
```
jonkoops, line 53: Could you explain this 'debug' mode? Why not simply use the existing mode we already provide to Webpack?

jonkoops (changes requested):

jonkoops:
I would also be nice if we could inform the user when they loaded an older version of the site and a new one is available. Something much like this: https://deanhume.com/displaying-a-new-version-available-progressive-web-app/

chrisvanmook:
I'm going to check this out and test this thoroughly tomorrow or monday

index.ejs
```diff
@@ -72,14 +72,6 @@
       </span>
     </div>
 
-    <% for (let key in htmlWebpackPlugin.options.modernScripts) { %>
```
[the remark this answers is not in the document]
janjaap: Because Webpack injects the async scripts itself.

webpack.dev.ts
janjaap: This settings means that changes to files in the project's root folder, and not only the `src` folder, will trigger a compilation.

webpack.prod.ts
janjaap: I added this, because it's handy to have debugger statements in production code when running the application in Docker locally. By default, debugger statements are removed from the production bundle, but not when the env var `DEBUG` has been set.

webpack.prod.ts
janjaap: That's an option that's only available in Webpack. We're still using 4.

webpack.prod.ts
janjaap: Because the existing mode for the Webpack production configuration is always 'production'. When working on the service worker configuration, it's useful to not have a minified version of that service worker. The value 'development', used here, can be anything, as long as it isn't 'production'.

webpack.prod.ts
janjaap: They're still valid Webpack 4 options.

src/store/store.js
```diff
@@ -16,7 +16,7 @@ import matomoMiddleware from './middleware/matomo/matomoMiddleware'
 import preserveUrlParametersMiddleware from './middleware/preserveUrlParametersMiddleware'
 import setQueriesFromStateMiddleware from './middleware/setQueriesFromStateMiddleware'
 import paramsRegistry from './params-registry'
-import './queryParameters'
+// import './queryParameters'
```
janjaap, line 19: I'll revert this change; this import definitely does something.

webpack.common.ts
janjaap: Good one; I removed the hashes from the JS chunks, because the service worker already takes care of hashing those files. I assumed that this was also the case for CSS files, but I'll verify that.

janjaap: Tested it and the `[name].css` format gets picked up the service worker as expect; changes are applied.

webpack.prod.ts
janjaap: Yeah, that is true. Generating two sets of chunks isn't picked up by the service worker. I'll see if I can make it work with both sets.

webpack.prod.ts
janjaap: Service worker takes care of hashing.

webpack.prod.ts
janjaap: Leaving the prop `splitChunks` out of the settings won't generate any chunks.

jonkoops (changes requested):

index.ejs
```diff
@@ -93,5 +93,12 @@
         }
       </script>
     <% } %>
+    <script>
+      if ('serviceWorker' in navigator && '${DEPLOY_ENV}' === 'production') {
```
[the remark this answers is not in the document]
janjaap: The value for `DEPLOY_ENV` is 'production' for both the acceptance and production environments, so that case is covered.

package.json
```diff
@@ -8,6 +8,7 @@
   },
   "version": "3.22.0",
   "scripts": {
+    "prebuild": "rimraf ./dist",
```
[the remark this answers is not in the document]
janjaap: Nice, I'll remove it. Good catch.

index.ejs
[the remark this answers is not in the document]
jonkoops: If I look at the log on the acceptance environment it states it is `acceptance`, so I do not think that the service worker will actually be loaded there:

image

src/store/store.js
jonkoops: @janjaap this has been resolved since?

index.ejs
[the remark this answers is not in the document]
janjaap: Ah, yes. I see. I was thrown off by the values that I found in the deploy config in Gitlab. Has been changed.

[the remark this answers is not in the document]
jonkoops: Yeah, this environment variable is a bit silly. I would like to get rid of it at some point.

webpack.prod.ts
```diff
@@ -22,22 +25,38 @@ export default merge(createConfig({ mode: 'production' }), {
...
             // Do not drop debugger statements, we might want to run a production build locally for testing.
             // Linting rules will ensure this never actually happens with true production images.
-            drop_debugger: false,
+            drop_debugger: !debugMode,
           },
         },
       }),
     ],
+    runtimeChunk: 'single',
     splitChunks: {
```
jonkoops, line 34: I am still not a fan of this automatic chunking for several reasons:

If I run a test on HTTP/1.1 I can see that a lot of the connections are stalling when I slow down the network in Chrome devtools. The screen-shot below shows the stalling time in gray:

image

This results in a slower initial page load of about 0.5s on the 'fast 3G' preset in Chrome dev tools.

Compare that to the fatter chunks that can all be downloaded in parallel:

image
 
Perhaps that HTTP/2 could fix this issue, but I would rather not include this change now. Also because it would only increase the surface area of possible missing chunks and thus errors.

webpack.prod.ts
```diff
@@ -11,8 +14,8 @@ export default merge(createConfig({ mode: 'production' }), {
     },
   },
   output: {
-    filename: '[name].[contenthash].js',
-    chunkFilename: '[name].[contenthash].js',
```
jonkoops, line 15: I would prefer to restore the asset fingerprinting logic in these file names. The reason being is that we now end up with unhashed chunks for small chunks such as the `ArticleDetailPage`:

image

This might look fine, but I am afraid that we might load a newer `ArticleDetailPage` chunk from an older bundle for the application code, which might cause errors.

jonkoops (changes requested):

webpack.prod.ts
jonkoops: This point is moot due to SW caching.

webpack.prod.ts
jonkoops: This will be solved by using HTTP/2 in the future.

jonkoops (approved):

jonkoops (approved):

PR #3664, opened 2020-11-24
[DI-1392] Remove legacy chunks

janjaap:
This PR removes the generation of backwards compatible JS chunks.


webpack.common.ts
```diff
@@ -122,15 +107,8 @@ export function createConfig(additionalOptions: CreateConfigOptions): Configurat
                 [
                   '@babel/preset-env',
                   {
-                    modules: false,
-                    ...(!options.legacy
```
jonkoops, line 126: I do not know exactly why I configured this as such but if we want to use ES2015+ methods in browsers that support modules we will still have to include the correct polyfills. It would be best to always set `useBuiltIns` and `corejs` to cover our bases here.

webpack.prod.ts
```diff
@@ -3,130 +3,45 @@ import TerserPlugin from 'terser-webpack-plugin'
...
+    filename: '[name].[hash].js',
+    chunkFilename: '[name].[chunkhash].js',
+  },
+  devtool: 'source-map',
+  optimization: {
+    minimizer: [
+      new TerserPlugin({
+        terserOptions: {
+          compress: {
+            drop_debugger: !debugMode,
```
jonkoops, line 25: We already have a linting rule to make sure that no debugging statements exist in production builds. Can we perhaps drop the `debugMode` entirely?

image

webpack.prod.ts
```diff
@@ -3,130 +3,45 @@ import TerserPlugin from 'terser-webpack-plugin'
...
+          sourceMap: true,
         },
+      }),
+    ],
+    namedChunks: true,
+    namedModules: true,
+    moduleIds: 'named',
+    chunkIds: 'named',
+    runtimeChunk: false,
+    splitChunks: {
```
jonkoops, line 36: Testing this locally `splitChunks` creates a **lot** of seperate files:
Screenshot from 2020-12-01 12-20-38

I think it would be best to use the WebPack defaults here which makes things a lot more manageable:
Screenshot from 2020-12-01 12-27-26

webpack.prod.ts
```diff
@@ -3,130 +3,45 @@ import TerserPlugin from 'terser-webpack-plugin'
...
+          compress: {
+            drop_debugger: !debugMode,
+          },
+          sourceMap: true,
         },
+      }),
+    ],
+    namedChunks: true,
+    namedModules: true,
+    moduleIds: 'named',
```
jonkoops, line 33: Considering that we want to be able to cache these files based on the file name in which case we would want to use the `deterministic` value as it will not change between compilations, from the [WebPack docs](https://webpack.js.org/configuration/optimization/#optimizationchunkids):

> 'deterministic' - Short numeric ids which will not be changing between compilation. Good for long term caching. Enabled by default for production mode.

The documentation also says the following:

> If the environment is development then optimization.chunkIds is set to 'named', while in production it is set to 'deterministic'

In my opinion we should not provide this value, and instead fall back on the more sensible defaults from WebPack.

jonkoops (changes requested):

package.json
```diff
@@ -163,26 +158,27 @@
...
     "jest-fetch-mock": "3.0.3",
     "jest-styled-components": "7.0.3",
+    "lint-staged": "10.5.2",
     "lodash.clonedeep": "^4.5.0",
     "mini-css-extract-plugin": "^1.3.1",
     "postcss": "^8.1.10",
     "postcss-loader": "^4.1.0",
     "prettier": "2.2.0",
-    "react-test-renderer": "17.0.1",
+    "prop-types": "15.7.2",
```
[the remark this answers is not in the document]
janjaap: Well, it shouldn't be used in production, so it makes sense to list it as a dev dependency. Just for good measure, because it makes no difference since we're not installing our dependencies with the `--production` option.

webpack.common.ts
janjaap: Good one, I'll change that.

webpack.prod.ts
janjaap: The point of `debugMode` is that, when running the application locally with Docker, we can still set debugger statements in the code that aren't removed. This helps with stepping through the production code.

webpack.prod.ts
janjaap: I don't agree; the Webpack defaults will throw everything in one chunk which is really big. This forces that chunk to be downloaded again with any change to the code. By generating more chunks, the chance of having to re-download a large part of the application on a new release, become a lot smaller and thus adding to the run-time and overall performance of the application.
Browsers can handle up to six requests simultaneously, so having as many smaller chunks sd with the change in this PR, is not an issue.

webpack.prod.ts
janjaap: `deterministic` Is specific to Webpack 5; we're still using 4. And more importantly, long-term caching will handled by the service worker.

package.json
```diff
@@ -85,6 +77,9 @@
   },
   "devDependencies": {
     "@babel/core": "^7.12.9",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.12.1",
```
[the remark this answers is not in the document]
janjaap: I added those because of errors during testing. I ran the tests and these errors don't pop up anymore 🤔 .

Anywho, I'll remove the deps.

.browserslistrc
```diff
@@ -2,5 +2,4 @@ last 2 Chrome versions
 last 2 Firefox versions
 last 2 Safari versions
 Firefox ESR
-IE 11
```
chrisvanmook, line 5: 🎉

chrisvanmook (approved):

PR #3665, opened 2020-11-24
(Chore) Allow localhost connections

janjaap:
This PR contains a change to the Nginx configuration so that connection on localhost can be made; important when setting the project locally with Docker.


janjaap:
Closing this PR; trying to solve it with `docker-compose` instead.

PR #3689, opened 2020-11-26
[DI-1392] nomodule support message

janjaap:
This PR contains a change to the public html file that will show a message about unsupported browsers. In order to not show the root loading indicator, its rendering is made dependent on module support.


index.ejs
```diff
@@ -1,83 +1,93 @@
...
-      <link rel="preload" href="<%= item %>" as="style">
-      <link rel="stylesheet" href="<%= item %>">
-    <% } %>
-
-    <style>
-      #root-loading-indicator {
-        display: flex;
-        padding: 2em 1em;
+
+<head>
```
jonkoops, line 4: Can we keep the indentation the same as before?

index.ejs
```diff
@@ -1,83 +1,93 @@
...
+        transform: rotate(360deg);
       }
+    }
+  </style>
 
-      @keyframes rotating-spinner {
-        0% {
-          transform: rotate(0deg);
-        }
+  <script nomodule>
```
jonkoops, line 67: The specs of the story specify that we want to do a redirect to a static page in this case and a nicer and styled message there. I'd recommend we serve a static HTML page from NGINX.

index.ejs
```diff
@@ -1,83 +1,93 @@
...
-        }
-      }
-    </style>
-  </head>
-  <body>
-    <div id="root" class="root">
+<body>
+  <div id="root" class="root"></div>
+
+  <script module>
```
jonkoops, line 75: If we redirect to a static page we can leave this script without a `module` attribute.

jonkoops (dismissed):

index.ejs
janjaap: I formatted the document in VS Code, so maybe we need to update the linter rules?

index.ejs
janjaap: I explored that possibility and I don't think it's feasible. In my opinion we have three options:
1. Use the HTML from data.amsterdam.nl
2. Use the HTML from www.amsterdam.nl
3. Don't use a frame at all

## Use the HTML from data.amsterdam.nl
Showing a message saying that we don't support older browsers would be best framed in the HTML that is already available at data.amsterdam.nl. However, doing that requires us to provide polyfills for those older browsers. If we don't, we cannot render the page. Providing polyfills for browsers that we don't want to provide polyfills for, is doing the exact opposite of what we want to achieve.

## Use the HTML from www.amsterdam.nl
We could use a static page from www.amsterdam.nl as the basis of the frame to show our support message in. The benefit is that we can reference all CSS and JS files on www.amsterdam.nl without having to worry about having to provide any polyfills (as in the previous option). However, the page will have a different layout from data.amsterdam.nl and we don't have control over the CSS nor the JS files and any changes in there might have a negative impact on our static page

## Don't use a frame at all
The benefit of only showing a plain message to the user is that there are no dependencies to take into account (polyfills, external CSS, JS, ...), we don't have to maintain another page in the application, the solution will always work, regardless of which browser the visitor uses and it doesn't take it tons of time to create it, unlike the options listed above.

All options have been discussed with UX and the third option was deemed most suitable for our situation.

If you have another take on this, I would definitely like to explore that.

index.ejs
jonkoops: We don't have linting rules for HTML files I believe so it will use the default formatter for VSCode. Let's not format things with IDE specific formatters if we have no linting set up for it though.

index.ejs
jonkoops: Thanks for the explanation, the original ticket was however quite clear on this. I also have to disagree with the options you've provided here.

We have an NGINX server in our application which could match a specific route and serve a piece of static HTML without issue. For example we could serve a file called `no-browser-support.html` from `https://data.amsterdam.nl/no-browser-support`. If the user goes to the main site we can detect the support for modules, and if it is not present call `window.location.assign('https://data.amsterdam.nl/no-browser-support')`.

It might also be a good idea to include Matomo on this page to be able to track how many users are using unsupported browsers.

index.ejs
janjaap: Fair enough, I'll revert this change.

chrisvanmook (approved):

chrisvanmook (approved):

chrisvanmook (approved):

index.ejs
janjaap: Very true. Removed.

PR #3706, opened 2020-11-30
(Chore) split nginx conf

janjaap:
This PR contains a split in Nginx configuration for development and production; some settings are required to be able to run the application locally and other settings shouldn't be in the production config.


chrisvanmook:
What could be the potential "danger" when including localhost in the CSP rules? I'm missing the sole purpose of why this pull-request is made.
Anyway, the downside of this is that we have to maintain 2 config files again... Perhaps instead of 2 main config files files, we could create 2 files (dev & pros, that will contain the CSP rules only), copy one of those in the Dockerfile and include that file in `nginx.conf`.

janjaap:
No danger, since the `nginx.dev.conf` file is only used locally. The production config is still intact. We need two config files, because the production config requires different settings and we shouldn't have local-specific settings in a production config file.

chrisvanmook:
> No danger, since the `nginx.dev.conf` file is only used locally. The production config is still intact. We need two config files, because the production config requires different settings and we shouldn't have local-specific settings in a production config file.

I see. Then I'd prefer to create separate files for only CSP rules

chrisvanmook (approved):

PR #3707, opened 2020-11-30
BetaMap: Fixed several bugs in drawtool, changed URl structure and ignore certain redux middleware when on betamap

chrisvanmook:
In this PR I've done the following things:
- Make sure "Detail", "Geosearch"and "Panorama" pages follow the same structure as the legacy URL, except instead of beginning with "/data" we use "/kaart" for now, so the legacy pages can still exist
- URLs containing "/kaart" (so if user is on beta map) it will skip certain redux middleware, causing loads of side-effects
- Added a warning to `useParam` hook
- Fixed several bugs regarding the draw-tool


src/app/components/PanoramaViewer/PanoramaViewerMarker.tsx
```diff
@@ -28,7 +28,7 @@ const PanoramaViewerMarker: React.FC<MarkerProps> = ({ location, setLocation })
   useMapCenterToMarker(location)
 
   useMapEvents({
-    click: ({ latlng }) => setLocation(latlng),
+    click: ({ latlng }) => setLocation && setLocation(latlng),
```
janjaap, line 31: According to the Leaflet docs, the `latlng` object, has a method `isValid()`. Might be good to check that as well before calling `setLocation`?

src/app/pages/MapPage/MapControls.tsx
```diff
@@ -68,7 +70,14 @@ const MapControls: React.FC<Props> = ({
...
           }
-          topRight={!panoActive && <DrawTool setCurrentOverlay={setCurrentOverlay} />}
+          topRight={
+            !panoActive && (
+              <>
+                <DrawToolOpenButton onClick={() => setShowDrawTool(true)} />
+                {showDrawTool && <DrawTool setCurrentOverlay={setCurrentOverlay} />}
+              </>
+            )
+          }
```
janjaap, line 80: Add a test to cover this change?

src/app/pages/MapPage/MapControls.tsx
```diff
@@ -87,7 +96,20 @@ const MapControls: React.FC<Props> = ({
...
+                  <DrawToolOpenButton
+                    onClick={() => {
+                      setShowDrawTool(true)
+                    }}
+                  />
+                )}
+                {showDrawTool && <DrawTool setCurrentOverlay={setCurrentOverlay} />}
+              </>
+            )
+          }
```
janjaap, line 112: Same here, add a test?

src/app/pages/MapPage/MapPage.tsx
```diff
@@ -53,24 +53,14 @@ export const MAP_PANEL_SNAP_POSITIONS: PositionPerSnapPoint = {
...
-// Todo: get ID's from request
-const PANO_LAYERS = [
-  'pano-pano2020bi',
-  'pano-pano2019bi',
-  'pano-pano2018bi',
-  'pano-pano2017bi',
-  'pano-pano2016bi',
-]
-
 const MapPage: React.FC = () => {
```
janjaap, line 56: This component seems quite complex. Can you add a couple of tests to cover its functionality?

src/app/pages/MapPage/MapPanelContent.tsx
```diff
@@ -42,18 +43,36 @@ const MapPanelContent: React.FC<MapPanelContentProps> = ({ setCurrentOverlay, cu
...
+          ]}
+          exact
+        >
+          <DrawResults
+            {...{
+              currentOverlay,
+            }}
+          />
+        </Route>
+      </Switch>
```
janjaap, line 67: Shouldn't routing be happening at the highest level possible in the application?

src/app/pages/MapPage/detail/DetailPanel.tsx
```diff
@@ -144,13 +148,24 @@ const DetailPanel: FunctionComponent<DetailPanelProps> = ({ detailUrl }) => {
...
         <MapPanelContent
           title={value?.data.subTitle || 'Detailweergave'}
           subTitle={value?.data.title || ''}
-          onClose={() => setDetailUrl(null)}
+          onClose={() => {
+            history.push({
+              pathname: routing.dataSearchGeo.path,
+              search: buildQueryString([[panoParam, null]]),
+            })
+          }}
```
janjaap, line 168: Add test cases for these changes?

src/app/pages/MapPage/detail/buildDetailUrl.ts
```diff
@@ -1,22 +1,18 @@
...
-  return window.location.pathname === '/kaart' || window.location.pathname === '/kaart/'
-    ? {
-        pathname: routing.map.path,
-        search: buildParamQuery(detailUrlParam, `${type}/${subtype}/${id}`).toString(),
-      }
-    : {
-        pathname: generatePath(routing.dataDetail.path, { type, subtype, id: `id${id}` }),
-        search: window.location.search,
-      }
+  // Todo: AfterBeta: this can be removed
```
janjaap, line 8: Can the whole function body be removed or just line 10?

src/app/pages/MapPage/draw/DrawResults.tsx
```diff
@@ -208,6 +214,10 @@ const DrawResults: React.FC<Props> = ({ currentOverlay }) => {
       stackOrder={currentOverlay === Overlay.Results ? 2 : 1}
       onClose={() => {
         setShowDrawTool(false)
+        history.push({
+          pathname: routing.dataSearchGeo_TEMP.path,
+          search: buildQueryString(undefined, [polylinesParam, polygonsParam]),
+        })
```
janjaap, line 220: Cover with a test case?

src/app/pages/MapPage/draw/DrawTool.tsx
```diff
@@ -79,11 +84,16 @@ const DrawTool: React.FC<DrawToolProps> = ({ setCurrentOverlay }) => {
...
+
+  const [initialDrawnItems, setInitialDrawnItems] = useState<ExtendedLayer[]>([])
 
-  const [initialDrawnItems, setInitialDrawnItems] = useState([
-    ...polygons.map((drawing) => createPolyLayer(drawing)),
-    ...polylines.map((drawing) => createPolyLayer(drawing, true)),
-  ])
+  // We needs refs here for leaflet event handlers
+  const polygonsRef = useRef(polygons)
+  polygonsRef.current = polygons
```
janjaap, line 93: Wouldn't this also do the trick?
```jsx
const polygonsRef = useRef(polygons)
```
Instead of creating a ref with a specific value and then setting that value again.

src/app/pages/MapPage/draw/DrawTool.tsx
```diff
@@ -108,30 +118,64 @@ const DrawTool: React.FC<DrawToolProps> = ({ setCurrentOverlay }) => {
     }
   }, [])
 
+  const updateShapes = (shape: { polygons?: PolyDrawing[]; polylines?: PolyDrawing[] }) => {
+    window.localStorage.setItem('polygons', JSON.stringify(shape.polygons))
+    window.localStorage.setItem('polylines', JSON.stringify(shape.polylines))
```
janjaap, line 123: According to [caniuse.com](https://caniuse.com/namevalue-storage):
> In private browsing mode, Safari and iOS Safari up to including version 10.x as well as the Android browser (not include Chrome for Android) do not support setting sessionStorage or localStorage.
Why is it necessary to store these values in localStorage? Isn't it possible to use the URL params for that, like it is done for a lot of other occasions in the application?

src/app/pages/MapPage/draw/DrawTool.tsx
```diff
@@ -207,72 +265,19 @@ const DrawTool: React.FC<DrawToolProps> = ({ setCurrentOverlay }) => {
...
-      onToggle={setShowDrawTool}
-      drawnItems={initialDrawnItems && initialDrawnItems}
+      isOpen
+      onClose={() => {
+        setShowDrawTool(false)
+        history.push({
+          pathname: routing.dataSearchGeo_TEMP.path,
+          search: buildQueryString(undefined, [polylinesParam, polygonsParam]),
+        })
+      }}
```
janjaap, line 279: Shouldn't pushing a new state to the history be the responsibility of the component that renders the draw tool instead of the draw tool itself?

src/app/pages/MapPage/draw/DrawTool.tsx
```diff
@@ -68,7 +73,7 @@ export interface DrawToolProps {
 
 const DrawTool: React.FC<DrawToolProps> = ({ setCurrentOverlay }) => {
```
janjaap, line 74: This component doesn't have any tests. Please increase the coverage on it. There is a lot happening in this component that becomes more understandable by adding tests.

src/app/pages/MapPage/map-search/MapSearchMarker.tsx
```diff
@@ -1,57 +1,64 @@
...
-    const layers = activeOverlays
-      .filter((overlay) => !!overlay.layer.detailUrl)
-      .map((overlay) => overlay.layer)
+  const legendLeafletLayersRef = useRef<Overlay[]>(legendLeafletLayers)
 
-    if (layers.length === 0) {
-      setLocation(e.latlng)
-      return
+  useEffect(() => {
+    if (legendLeafletLayers && legendLeafletLayersRef) {
```
janjaap, line 26: Isn't `legendLeafletLayersRef` always truthy? The call to `useRef` always returns an object.

src/app/pages/MapPage/map-search/MapSearchMarker.tsx
```diff
@@ -1,57 +1,64 @@
...
+      legendLeafletLayersRef.current = legendLeafletLayers
     }
+  }, [legendLeafletLayers])
+
+  async function handleMapClick(e: LeafletMouseEvent) {
+    const layers = legendLeafletLayersRef.current
+      .filter((overlay) => !!overlay.layer.detailUrl)
```
janjaap, line 33: Double negation isn't necessary in case of validating a string for its truthy value

src/app/pages/MapPage/map-search/MapSearchResults.tsx
```diff
@@ -77,26 +76,27 @@ const StyledPanoramaPreview = styled(PanoramaPreview)`
...
-    [location.lat, location.lng, user],
+      location
+        ? mapSearch(
+            {
+              latitude: location.lat,
+              longitude: location.lng,
+            },
+            user,
+          )
+        : Promise.reject(),
```
janjaap, line 95: Please cover this by a test case.

src/app/pages/MapPage/map-search/PanoramaPreview.tsx
```diff
@@ -69,6 +71,14 @@ const PanoramaPreview: React.FC<PanoramaPreviewProps> = ({
   radius,
   ...otherProps
 }) => {
+  const [activeLayers] = useParam(mapLayersParam)
+  const activeLayersWithoutPano = useMemo(
+    () => activeLayers.filter((id) => !PANO_LAYERS.includes(id)),
+    [],
+  )
+
+  const newLayers = [...activeLayersWithoutPano, ...PANO_LAYERS]
```
janjaap, line 80: Shouldn't this value be a `useMemo` value instead of `activeLayersWithoutPano`?

src/app/pages/MapPage/map-search/PanoramaPreview.tsx
```diff
@@ -78,20 +88,21 @@ const PanoramaPreview: React.FC<PanoramaPreviewProps> = ({
         aspect,
         radius,
       }),
-    [location.lat, location.lng, width, fov, horizon, aspect, radius],
+    [location],
```
janjaap, line 91: The removes dependencies are still used. Why is it necessary to remove them from the list?

src/app/pages/MapPage/map-search/PanoramaPreview.tsx
```diff
@@ -105,25 +116,36 @@ function renderResult(
...
+    : toPanoramaAndPreserveQuery(result?.value?.id, result?.value?.heading, legacyReference)
+  const linkComponent = window.location.pathname.includes('kaart') ? RouterLink : LegacyLink
+
   return (
     <>
       <PreviewImage src={result.value.url} alt="Voorvertoning van panoramabeeld" />
       {/*
       // @ts-ignore */}
-      <PreviewLink forwardedAs={linkComponent} to={link} inList>
+      <PreviewLink forwardedAs={linkComponent} to={to} inList>
```
janjaap, line 148: Please cover the value of `to` with a test case

src/app/utils/useParam.ts
```diff
@@ -34,7 +34,18 @@ const useParam = <T>(urlParam: UrlParam<T>): [T, SetValueFn<T>] => {
...
+      // eslint-disable-next-line no-console
+      console.warn(
+        "There's a mismatch between window.location and the location from the useLocation hook from react-router. This sometimes happens when calling the 'setValue' inside a non-react event-handler that is not updated. Please be sure the component is updated via react-router",
+        'Tried to update parameter with key: "',
+        urlParam.name,
+        '"',
+      )
+
+      return undefined
+    }
```
janjaap, line 47: Shouldn't this throw instead of show a warning? Seems serious enough to break.

src/store/middleware/addMetaToRoutesMiddleware.js
```diff
@@ -3,6 +3,9 @@ import paramsRegistry from '../params-registry'
 let nrOfRoutesDispatched = 0
 
 const addMetaToRoutesMiddleware = ({ getState }) => (next) => (action) => {
+  if (window.location.pathname.includes('kaart')) {
+    return next(action)
+  }
```
janjaap, line 8: Add a test? If you do, then at least the test will break if this condition is not removed when transitioning from old to new map.

src/store/middleware/preserveUrlParametersMiddleware.js
```diff
@@ -4,6 +4,10 @@ import ParamsRegistry from '../params-registry/paramRegistry'
 
 const preserveUrlParametersMiddleware = () => (next) => (action) => {
   let nextAction = action
+
+  if (window.location.pathname.includes('kaart')) {
+    return next(nextAction)
+  }
```
janjaap, line 10: Same, please add a test

src/store/middleware/setQueriesFromStateMiddleware.js
```diff
@@ -2,6 +2,9 @@ import paramsRegistry from '../params-registry'
 import { getLocationType } from '../redux-first-router/selectors'
 
 const urlParamsMiddleWare = ({ getState }) => (next) => (action) => {
+  if (window.location.pathname.includes('kaart')) {
+    return next(action)
+  }
```
janjaap, line 7: Add a test, please.

janjaap (changes requested):

src/app/components/PanoramaViewer/PanoramaViewerMarker.tsx
chrisvanmook: If I look into the type, only `LatLngBounds` have this. We don't really need to worry here anyway since it's typed

jonkoops (changes requested):

src/app/pages/MapPage/MapPanelContent.tsx
chrisvanmook: Why should that be? I think the whole idea of RR is that you can render components conditionally based on the pathname, not per se on the highest level. In this case our view is the Map, but the content in the sidepanel can vary from "dataselectie", "geozoeken" or "detail-pages".

janjaap: For no reason in particular, was just wondering what the approach is in the application.

src/app/pages/MapPage/detail/buildDetailUrl.ts
chrisvanmook: just the statement, will update the comment

src/app/pages/MapPage/draw/DrawTool.tsx
chrisvanmook: Thought that would work as well, but unfortunately it didn't... IMO we should find a solution in `react-map` and `ARM` to fix this, like passing a dependency array to every event so it knows when to remove and add the eventlisteners again.

src/app/pages/MapPage/draw/DrawTool.tsx
chrisvanmook: We also store them in the URL, the localstorage is a nice-to-have improvement in case users want to go back to their drawing after closing the tool and search for other stuff

package.json
```diff
@@ -50,7 +50,7 @@
     "isomorphic-fetch": "3.0.0",
     "json-api-normalize": "1.1.5",
     "leaflet": "1.7.1",
-    "leaflet-draw": "1.0.0",
+    "leaflet-draw": "1.0.4",
```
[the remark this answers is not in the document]
chrisvanmook: It will, I'll remove it when this PR is approved. Meanwhile I'll put a label on it

src/app/utils/buildQueryString.ts
```diff
@@ -1,10 +1,13 @@
 import encodeParam from './encodeParam'
 import { UrlParam } from './useParam'
 
-export default function buildQueryString<T>(params: [UrlParam<T>, T][]) {
+export default function buildQueryString<T>(
+  paramsToAdd?: [UrlParam<T>, T][],
+  paramsToEmit?: Array<UrlParam<any>>,
```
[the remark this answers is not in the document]
chrisvanmook: Lol haha yes :D

src/app/components/PanoramaViewer/PanoramaViewer.tsx
```diff
@@ -99,16 +113,29 @@ const PanoramaViewer: React.FC = () => {
     }
   }, [marzipanoViewer, location, panoTag])
 
+  const activeLayersWithoutPano = useMemo(
+    () => activeLayers.filter((id) => !PANO_LAYERS.includes(id)),
+    [],
```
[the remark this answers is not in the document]
chrisvanmook: `activeLayers` you mean then?

src/app/pages/MapPage/map-search/MapSearchMarker.tsx
chrisvanmook: Good point! will remove

src/app/pages/MapPage/draw/DrawTool.tsx
chrisvanmook: Regarding the size of this pull-request, I'm going to create another issue in Jira for this.

src/app/pages/MapPage/map-search/MapSearchMarker.tsx
chrisvanmook: indeed! will remove

src/app/components/PanoramaViewer/PanoramaViewer.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    const closeButton = await findByTestId('panoramaViewerCloseButton')
+    expect(mockPush).not.toHaveBeenCalled()
+
+    fireEvent.click(closeButton)
+
+    expect(mockPush).toHaveBeenCalledWith({
+      pathname: 'some-url/',
+      search: 'randomParam=%22should-stay%22',
+    })
```
janjaap, line 30: Would be better to check the value of `window.location`, because that's the result that we're after when pushing a new item on the history stack. Doesn't really matter that we accomplish that by using `react-router-dom` or some other means.

src/app/pages/MapPage/MapControls.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+          panoActive={false}
+        />,
+        {
+          setShowDrawTool: setShowDrawToolMock,
+        },
+      ),
+    )
+    expect(setShowDrawToolMock).not.toHaveBeenCalled()
+    const openDrawtoolButton = getByTestId('drawtoolOpenButton')
+    fireEvent.click(openDrawtoolButton)
```
janjaap, line 72: Assert that `setShowDrawToolMock` hasn't been call prior to clicking the button.

src/app/pages/MapPage/MapPage.test.tsx
```diff
@@ -1,85 +1,63 @@
...
-    dispatchEvent: jest.fn(),
-  })),
-})
+jest.mock('react-router-dom', () => ({
+  // @ts-ignore
+  ...jest.requireActual('react-router-dom'),
+  useLocation: () => ({
+    search: 'pano={"foo":"bar"}',
+  }),
+}))
```
janjaap, line 30: Instead of having to rely on `react-router-dom` (we might something else in the future), you can also directly change the value of `window.location`. You'd probably have to make that object writable, like so:

```javascript
  Object.defineProperties(global, {
    location: {
      writable: true,
      value: global.location,
    },
  })

```

src/map/components/search-results/MapSearchResults.test.jsx
```diff
@@ -10,11 +10,16 @@ jest.mock('redux-first-router-link', () => ({
...
 jest.mock('react-redux', () => ({
   useSelector: jest.fn(() => []),
 }))
 
+jest.mock('../../../app/utils/useGetLegacyPanoramaPreview', () => () => ({
+  panoramaUrl: '/foo',
+  link: '/link',
+  linkComponent: null,
+}))
+
```
janjaap, line 22: Why is it necessary to mock this module?

src/app/pages/MapPage/map-search/PanoramaPreview.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+jest.mock('../../../../api/panorama/thumbnail')
+
+describe('PanoramaPreview', () => {
+  beforeEach(() => {
+    // @ts-ignore
+    getPanoramaThumbnail.mockImplementation(() => Promise.resolve({ url: 'example.com' }))
+  })
+  it('should build a link including current parameters, panorama parameters and layers for panorama', async () => {
+    const { container } = render(withAppContext(<PanoramaPreview location={{ lat: 1, lng: 2 }} />))
+    await waitFor(() => {
```
janjaap, line 25: Instead of wrapping this, you can also await the finding of an element by testid:
```jsx
await findByTestId('panoramaPreview')
```

test/fixtures/gebieden/stadsdeel.json
```diff
@@ -0,0 +1 @@
...8.378],[124060.986,487004.515],[124060.081,487039.836],[124064.088,487044.3],[124064.916,487044.375],[124062.897,487059.897],[124061.116,487073.194],[124060.096,487073.058],[124059.808,487075.082],[124056.662,487075.094],[124054.473,487076.235],[124051.881,487080.436],[124051.287,487099.625],[124042.025,487141.476],[124033.941,487166.65],[124021.653,487188.339],[124021.389,487192.022],[124022.194,487193.985],[124026.598,487197.909],[124017.349,487217.675],[124011.968,487217.226],[124006.956,487220.104],[123998.093,487239.392],[123982.886,487261.032],[123961.484,487287.469],[123944.286,487305.869],[123924.101,487324.852],[123901.815,487342.416],[123877.388,487359.121],[123854.637,487372.619],[123823.894,487387.201],[123766.046,487410.511],[123722.326,487429.979]]]]},"buurten":{"count":70,"href":"https://acc.api.data.amsterdam.nl/gebieden/buurt/?format=json&stadsdeel=03630000000018"},"buurtcombinaties":{"count":10,"href":"https://acc.api.data.amsterdam.nl/gebieden/buurtcombinatie/?format=json&stadsdeel=03630000000018"},"gebiedsgerichtwerken":{"count":2,"href":"https://acc.api.data.amsterdam.nl/gebieden/gebiedsgerichtwerken/?format=json&stadsdeel=03630000000018"},"dataset":"gebieden"}
```
janjaap, line 1: Can you move this file to the `src/api/gebieden/stadsdeel` folder? There are already examples in the `src/api` folder. Would also be nice if you can add a `types.ts` and have the fixture exported by that type so that tests can make use of those typings.

src/app/pages/MapPage/map-search/PanoramaPreview.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+      const link = container.querySelector('a')
```
janjaap, line 26: You could also have clicked the link and have verified the value of `window.location`, but checking the element's params also does the trick

src/app/pages/MapPage/map-search/PanoramaPreview.test.tsx
chrisvanmook: In this case I want to select the link to the panorama, that doesn't have a testid. So we could either do this or add a data-testid to the link itself

src/map/components/search-results/MapSearchResults.test.jsx
chrisvanmook: I've mocked this for now, as I would completely have to refactor this test if I wouldn't. IMO it's not worth the time and effort to do so, as this component (including the mocked hook) will be completely removed after releasing the new Map module

src/app/pages/MapPage/MapPage.test.tsx
chrisvanmook: I do agree, but I cannot get this to work. If I log `window.location` in the component it show the right value that I've set in the test. However, the value coming from `useLocation` is not equal to that value. No idea why...

src/app/pages/MapPage/MapControls.test.tsx
chrisvanmook: I already do that on line 70 :)

src/app/components/PanoramaViewer/PanoramaViewer.test.tsx
chrisvanmook: Like said here https://github.com/Amsterdam/atlas/pull/3707#discussion_r536084744, I cannot get this to work for now. Let's discuss this next week :)

src/app/pages/MapPage/map-search/PanoramaPreview.test.tsx
chrisvanmook: Could've done that as well, not really making any difference (IMHO a bit pointless to test the integrity of the router)

src/app/utils/useParam.ts
chrisvanmook: won't be resolved when remove the dual setup. I think this is the case: it happens when the values from `useParam` are used in a non-react eventhandler, and the value of eg. `setLocation` is still the same (and holds an 'old' value of `location`, coming from `useLocation`), even after a re-render of the component. Happens a lot with leaflet events. We should make a fix for this in the ARM and react-maps repo.

src/app/pages/MapPage/draw/DrawTool.tsx
chrisvanmook: Unfortunately we decided to keep the current interaction, which works as follows:
- using the draw measure tool, does not change the location, only updates the parameter in the URL
- using the draw polygon tool, DOES trigger a location change: the user should navigate to so-called 'Dataselection' page (like `/data/bag/adressen`) because the user "selected" an area on the Map.

src/app/pages/MapPage/map-search/PanoramaPreview.tsx
chrisvanmook: Because it's not necessary to update this every time, since it's just a preview of the Panorama. only when the location changes it should update

src/store/middleware/addMetaToRoutesMiddleware.js
janjaap: In the coverage repo I can see that the branch on line 10 isn't fully covered. Can you have a look?

src/store/middleware/preserveUrlParametersMiddleware.js
janjaap: The branches on lines 18 and 20 aren't covered by your tests. Can you also have a look there?

src/api/gebieden/stadsdeel.json
```diff
@@ -0,0 +1 @@
...8.378],[124060.986,487004.515],[124060.081,487039.836],[124064.088,487044.3],[124064.916,487044.375],[124062.897,487059.897],[124061.116,487073.194],[124060.096,487073.058],[124059.808,487075.082],[124056.662,487075.094],[124054.473,487076.235],[124051.881,487080.436],[124051.287,487099.625],[124042.025,487141.476],[124033.941,487166.65],[124021.653,487188.339],[124021.389,487192.022],[124022.194,487193.985],[124026.598,487197.909],[124017.349,487217.675],[124011.968,487217.226],[124006.956,487220.104],[123998.093,487239.392],[123982.886,487261.032],[123961.484,487287.469],[123944.286,487305.869],[123924.101,487324.852],[123901.815,487342.416],[123877.388,487359.121],[123854.637,487372.619],[123823.894,487387.201],[123766.046,487410.511],[123722.326,487429.979]]]]},"buurten":{"count":70,"href":"https://acc.api.data.amsterdam.nl/gebieden/buurt/?format=json&stadsdeel=03630000000018"},"buurtcombinaties":{"count":10,"href":"https://acc.api.data.amsterdam.nl/gebieden/buurtcombinatie/?format=json&stadsdeel=03630000000018"},"gebiedsgerichtwerken":{"count":2,"href":"https://acc.api.data.amsterdam.nl/gebieden/gebiedsgerichtwerken/?format=json&stadsdeel=03630000000018"},"dataset":"gebieden"}
```
janjaap, line 1: Can you add types for this file? This will help catching edge cases in tests.

src/api/gebieden/stadsdeel/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  dataset: string
+}
+
+export interface Stadsdeel {
+  _links: Links
+  _display: string
+  stadsdeelidentificatie: string
+  code: string
+  /* eslint-disable camelcase */
+  date_modified: string
```
chrisvanmook, line 17: @jonkoops this should be of type `Date`, however in `fixture.ts` it shows `type 'string' is not comparable to type 'Date'. json file`. Any idea how to fix this?

src/app/pages/MapPage/detail/DetailPanel.test.tsx
```diff
@@ -3,7 +3,7 @@ import { fireEvent, render } from '@testing-library/react'
 import fetch from 'jest-fetch-mock'
 import withMapContext from '../../../utils/withMapContext'
 import DetailPanel from './DetailPanel'
-import stadsdeelResponse from '../../../../api/gebieden/stadsdeel.json'
+import stadsdeelResponse from '../../../../api/gebieden/stadsdeel/stadsdeel.json'
```
janjaap, line 6: If you import from '../../../../api/gebieden/stadsdeel/fixture' you should get typing support

janjaap (approved):

PR #3749, opened 2020-12-10
Move runtime dependencies from devDependencies

jonkoops:


package.json
```diff
@@ -33,11 +33,14 @@
     "@amsterdam/asc-assets": "^0.26.2",
     "@amsterdam/asc-ui": "^0.26.1",
     "@amsterdam/react-maps": "^0.10.0",
+    "@babel/runtime": "^7.12.5",
```
janjaap, line 36: What difference do these changes make?

janjaap: `@babel/runtime` Isn't used in this repo

jonkoops: It's not imported directly, but it is used by [@babel/babel-plugin-transform-runtime](https://babeljs.io/docs/en/babel-plugin-transform-runtime).

jonkoops: > What difference does moving dependencies make?

These dependencies are used at runtime since they are part of the bundle, thus they should be included as a `dependency` not a `devDependency`. The latter is only meant for dependencies that are used during building and testing or related scripts.

janjaap: If we would have installed dependencies with the `--production` option on the command line, then moving any dependencies from runtime dependencies to dev dependencies (or the other way around) would make sense. In the current situation, in the current build pipeline, moving dependencies doesn't have any effect, whatsoever.

chrisvanmook (approved):

PR #3750, opened 2020-12-10
Use WebPack defaults instead of custom splitChunks

jonkoops:


webpack.prod.ts
```diff
@@ -32,16 +32,5 @@ export default merge(createConfig({ mode: 'production' }), {
     namedModules: true,
     moduleIds: 'named',
     chunkIds: 'named',
-    runtimeChunk: false,
```
jonkoops, line 35: We do not support HTTP/2 at the moment so having a lot of files is not good for performance.

PR #3761, opened 2020-12-14
Wrapped DatasetDetailPage in PromiseResult Component

chrisvanmook:


src/app/utils/usePromise.ts
```diff
@@ -22,7 +25,7 @@ export interface PromiseFulfilledResult<T> {
 
 export interface PromiseRejectedResult {
   status: PromiseStatus.Rejected
-  error: any
+  error: PromiseResultError
```
jonkoops, line 28: Probably best to just use the `Error` class here as a value, since all errors in our system extend `Error` we can still use `instanceof` to handle custom errors such as `AuthError`. Otherwise we have to specify every possible error in the system.

jonkoops (changes requested):

src/app/utils/usePromise.ts
chrisvanmook: Nice, I thought that would give errors when checking on `error.code`, but we also use `error instanceof AuthError` for example and it still works 👍

jonkoops (approved):

PR #3763, opened 2020-12-14
Added BAG v1 fixtures

chrisvanmook:


src/api/types.ts
```diff
@@ -1,15 +1,28 @@
+/* eslint-disable camelcase */
 export interface Links {
```
jonkoops, line 2: I believe we already have an existing type for hypermedia links in the root of the `api` dir.

src/api/types.ts
```diff
@@ -1,15 +1,28 @@
...
 }
 
-export interface APIReference {
-  count: number
+export interface SmallAPIReference {
+  count?: number
   href: string
 }
 
 export interface Geometrie {
```
jonkoops, line 13: This can be replaced with `Geometry` from the `geojson` package.

src/api/v1/bag/ligplaats/types.ts
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable camelcase */
+import { SmallAPIReference, APIReference, Geometrie } from '../../../types'
+
+export interface Root extends APIReference {
+  ligplaatsidentificatie: string
+  date_modified: string
+  document_mutatie: string
+  document_nummer: string
+  begin_geldigheid: string
+  einde_geldigheid: any
```
jonkoops, line 10: Why do we have `any` as a type here?

jonkoops (approved):

src/api/v1/bag/ligplaats/types.ts
chrisvanmook: Value in the json is `null`

PR #3801, opened 2021-01-02
[DI2-59] Replace Redux router links with React router

jonkoops:


chrisvanmook (approved):

src/app/links.ts
```diff
@@ -0,0 +1,52 @@
+import { LocationDescriptor } from 'history'
+import { generatePath } from 'react-router-dom'
+import { ViewMode } from '../shared/ducks/ui/ui'
+import parameters from '../store/parameters'
+import { routing } from './routes'
+
+export const toConstructionFile = (
```
janjaap, line 7: The name of the functions in this module are confusing; they suggest performing an action, but instead return an object that can be object as the value of the `to` prop on a `Link` component. Maybe `getConstructionFileLocation`?

src/app/links.test.ts
```diff
@@ -0,0 +1,41 @@
+import { generatePath } from 'react-router-dom'
+import { toAddresses, toCadastralObjects, toConstructionFile, toEstablishments } from './links'
+import { routing } from './routes'
+
+describe('toConstructionFile', () => {
+  it('creates a link', () => {
```
janjaap, line 6: Doesn't create a link, but return location properties that can be used to construct a link from.

janjaap (changes requested):
There's about twenty other components that make use of `redux-first-router-link`. Will those be tackled in this PR as well?

jonkoops:
> There's about twenty other components that make use of `redux-first-router-link`. Will those be tackled in this PR as well?

This was originally the plan, but we decided to split this up into multiple separate tasks so those will come as separate PRs.

src/app/pages/MapPage/draw/DrawResults.tsx
```diff
@@ -99,7 +99,7 @@ const StyledLabel = styled(Label)`
   }
 `
 
-const TableRouterLink = styled(RouterLink)`
+const TableRouterLink = styled(ReactRouterLink)`
```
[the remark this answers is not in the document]
jonkoops: That import cannot be removed yet as it is still used for some code (see line 45-48).

```ts
const ResultLink = styled(RouterLink)`
  width: 100%;
  margin-bottom: ${themeSpacing(2)};
`
```

src/app/links.ts
jonkoops: I agree that the naming is not fantastic but I think it makes sense in the context it's used. I'd like to keep it like this for now, we can always rename them all at the end.

src/app/links.test.ts
jonkoops: Changed this to 'location descriptor' like the type it returns.

PR #3804, opened 2021-01-03
Convert header code to TypeScript

jonkoops:


src/app/components/Header/HeaderMenu.tsx
```diff
@@ -35,17 +38,38 @@ const components = {
...
+export interface HeaderMenuProps {
+  type: keyof typeof components
+}
+
+const HeaderMenu: FunctionComponent<HeaderMenuProps & ComponentProps<typeof StyledMenuInline>> = ({
+  type,
+  ...otherProps
+}) => {
+  const dispatch = useDispatch()
+  const user = useSelector(getUser)
```
janjaap, line 55: This var isn't used, is it?

src/app/components/Header/HeaderMenu.tsx
```diff
@@ -58,14 +82,15 @@ const HeaderMenu = ({ type, login, logout, user, showFeedbackForm, ...props }) =
               trackEvent({
                 category: 'navigation',
                 action: 'main-menu',
-                name: title,
+                name: title ?? undefined,
```
janjaap, line 85: Shouldn't really matter if any of `null` or `undefined` is passed.

src/app/components/Header/HeaderMenu.tsx
```diff
@@ -58,14 +82,15 @@ const HeaderMenu = ({ type, login, logout, user, showFeedbackForm, ...props }) =
...
               })
               dropFocus()
             }}
             as={RouterLink}
             iconLeft={<ChevronRight />}
             key={id}
-            title={title}
+            title={title ?? undefined}
```
janjaap, line 92: This value shouldn't be an attr on the element since the element's text content is the same. See https://developer.paciellogroup.com/blog/2010/11/using-the-html-title-attribute/ for reference.

src/app/components/Header/HeaderMenu.tsx
```diff
@@ -91,6 +116,7 @@ const HeaderMenu = ({ type, login, logout, user, showFeedbackForm, ...props }) =
                   as={RouterLink}
                   iconLeft={<ChevronRight />}
                   title={title}
```
janjaap, line 118: See previous comment; `title` attribute on links is discouraged.

src/app/components/Header/HeaderMenu.tsx
jonkoops: It's used further down.

src/app/components/Header/HeaderMenu.tsx
jonkoops: Yeah, it should not. But `trackEvent` expects this to be explicit.

src/app/components/Header/HeaderMenu.tsx
jonkoops: True, removed it.

src/app/components/Header/HeaderMenu.tsx
jonkoops: Removed

chrisvanmook (approved):

PR #3828, opened 2021-01-08
Convert Redux store configuration to TypeScript

jonkoops:


chrisvanmook (approved):
I think the global in eslintrc for `__REDUX....` can also be removed

PR #3833, opened 2021-01-11
[DI-1244] Server handlers

janjaap:
This PR provides a way to proxy requests to API endpoints by providing data fixtures that are served by a serviceworker process that captures requests when running tests. (For more information on the approach, read https://kentcdodds.com/blog/stop-mocking-fetch). It comes down to not checking with what fetch has returned with, because in nine out of ten cases, this is not important. More important is how the component under test handles the data (or lack thereof) and deals with error responses.

This has a couple of advantages:
- API responses are centralized
- Responses always return the same data
- Response fixtures can be typed
- Tests can be simplified


test/server-handlers.js
```diff
@@ -0,0 +1,55 @@
...
+  rest.get(typeaheadUrl, async (req, res, ctx) => {
+    const typeaheadFixture = require('../src/api/typeahead/typeahead.json')
+    const typeaheadAuthFixture = require('../src/api/typeahead/typeahead_auth.json')
+
+    const fixture = isAuthenticated() ? typeaheadAuthFixture : typeaheadFixture
+    return res(ctx.json(fixture))
+  }),
+
+  rest.get(iiifMetadataUrl, async (req, res, ctx) => {
+    const bouwdossierFixture = require('../src/api/iiif-metadata/bouwdossier/fixture').default
```
jonkoops, line 25: Why are we using `require` statements here? Is it not possible to import the JSON file with a regular import statement?

test/server-handlers.js
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable global-require */
```
jonkoops, line 1: All of these files are written in JavaScript, since we are using `ts-jest` would it not be possible to use TypeScript here instead?

jonkoops (approved):

chrisvanmook (approved):

jest.config.js
```diff
@@ -31,6 +31,8 @@ module.exports = {
   setupFiles: ['raf/polyfill', './test/setup-jest.js'],
   snapshotSerializers: ['enzyme-to-json/serializer'],
   setupFilesAfterEnv: ['./test/mocks.js'],
+  // will be uncommented once the base branch is ready for merge
+  // setupFilesAfterEnv: ['./test/mocks.js', './test/setup-env.js'],
```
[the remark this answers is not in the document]
janjaap: Will do in the next PR

test/server-handlers.js
janjaap: We can, but import statements should be placed at the module's scope and we don't need to import all the fixtures at once; only when we actually need them.

test/server-handlers.js
janjaap: Possibly, yes

PR #3851, opened 2021-01-13
[DI-1244] jest-fetch-mock replace

janjaap:
This PR removes the `jest-fetch-mock` dependency, since it conflicts with the service worker proxy


src/app/components/IIIFThumbnail/IIIFThumbnail.test.tsx
```diff
@@ -1,6 +1,9 @@
-import { cleanup, render } from '@testing-library/react'
+import { cleanup, render, screen } from '@testing-library/react'
 import { shallow } from 'enzyme'
+import * as path from 'path'
+import * as fs from 'fs'
```
jonkoops, line 4: I think we can use the default import here:
```suggestion
import path from 'path'
import fs from 'fs'
```

src/app/components/IIIFThumbnail/IIIFThumbnail.test.tsx
```diff
@@ -36,41 +39,50 @@ describe('IIIFThumbnail', () => {
...
     // Fetch returns ok = false, so render not found image
     expect(image.src).toContain('not_found_thumbnail.jpg')
   })
 
   it('should call the fetch method with the user token and display the image', async () => {
-    const mockFetch = (fetch as jest.Mock).mockReturnValueOnce(
-      Promise.resolve({
-        ok: true,
-        blob: () => Promise.resolve({}),
+    let request: any
```
jonkoops, line 57: Can we not use the `Request` type here instead?

src/app/pages/MapPage/detail/DetailPanel.test.tsx
```diff
@@ -23,15 +23,16 @@ jest.mock('react-router-dom', () => ({
   }),
 }))
 
-describe('DetailPanel', () => {
+// tests need to be (re)written for this (quite complex) component
```
jonkoops, line 26: When will we do this? Is there a ticket for it?

src/shared/services/auth/auth.test.js
```diff
@@ -24,7 +24,7 @@ jest.mock('./parseAccessToken')
 
 const notExpiredTimestamp = () => Math.floor(new Date().getTime() / 1000) + 1000
 
-describe('The auth service', () => {
+describe.skip('The auth service', () => {
```
jonkoops, line 27: We're no longer testing this? Is there a ticket to restore this?

jonkoops (changes requested):

src/app/pages/MapPage/detail/DetailPanel.test.tsx
janjaap: Shouldn't that be part of the story you were working on for the detail panels?

src/shared/services/auth/auth.test.js
janjaap: Working on those tests as we speak; will create a separate PR for that

src/app/components/IIIFThumbnail/IIIFThumbnail.test.tsx
janjaap: No, we can't. The `msw` dep doesn't have a `@types` package.

jonkoops (approved):

src/app/pages/MapPage/detail/DetailPanel.test.tsx
jonkoops: Yeah, just a bummer that we need to disable this now.

janjaap: Meh, there was only one test case in this suite and that even didn't test much 🤷🏻‍♂️

PR #3853, opened 2021-01-13
Converted cypress files to TS

chrisvanmook:


cypress/integration/map-layers_spec.ts
```diff
@@ -119,18 +119,10 @@ describe('Check if all map layers are visible when selected', () => {
     cy.contains('1943 (1:2500, Dienst der Publieke Werken)').should('be.visible')
     cy.contains('1985 (1:1000, Dienst der Publieke Werken)').should('be.visible')
     cy.get(MAP_LAYERS.checkboxHK1909).check({ force: true }).should('be.checked')
-    cy.url('contains', '**hist-pw1909%3A1&legenda=true')
```
chrisvanmook, line 122: Removed these as this wouldn't check anything in the first place (wrong use of `url`). Also the URL's would only be checked in specifically this case, which is a bit arbitrary and this logic should be checked in a unit test instead

cypress/integration/datasets_spec.ts
```diff
@@ -101,9 +103,15 @@ describe('datasets module', () => {
...
-          cy.log(datasetsNumber)
-          expect(parseInt(datasetsNumber, 10)).lessThan(parseInt(Cypress.env('countDatasets'), 10))
+          const datasetsNumber = $datasetsText.text().match(/\d+/)?.[0]
+          if (datasetsNumber) {
+            cy.log(datasetsNumber)
+            expect(parseInt(datasetsNumber, 10)).lessThan(
+              parseInt(Cypress.env('countDatasets'), 10),
+            )
+          } else {
+            throw new Error('Could not retrieve the number of datasets!')
```
janjaap, line 113: Will this error ever occur? Also, why are you throwing here and not on line 95? There's a similar check.

jonkoops (approved):

PR #3854, opened 2021-01-14
[DI2-64] Remove svg-loader

chrisvanmook:
I understand this PR is absolutely huge and therefore hard to review, my apologies for this 😄. Please note the following that might help you speed up the review:
- Of all the 227 files, I've deleted 113 (half of the PR). These are mostly unused `.scss` and `.svg` files and even an unused component. Also I've removed a few tests that would only contain testing snapshots. 
- I've replaced legacy components (with icons) with ASC, modified some css here and there and discussed these changes with our designer. He approved
- A lot of components are being used by our legacy Map module. Some (s)css changes perhaps might not be ideal, but it'll do the trick for now. I'm not going to spend more time on components that will going to be replaced soon.
- Use GH "Viewed" checkbox when you've reviewed a file! Very useful if you want to take a break and continue later 🤡 They will be unchecked when changes are pushed. IMHO a bit pointless to create several PR's for this one because it would still take the same amount of work to review...
- Just keep this in mind: these changes don't necessarily have "huge" impact. I barely have changed any logic.


src/app/components/ActiveFilters/ActiveFilters.tsx
```diff
@@ -12,28 +14,26 @@ export interface ActiveFiltersProps {
...
-          </li>
-        ))}
-      </ul>
-    </div>
+    <StyledUl>
+      {filters.map(({ slug, label, option }) => (
+        <li key={slug}>
+          <FilterTag
+            type="button"
+            title={`Filter verwijderen: ${label}: ${option}`}
```
janjaap, line 29: A `title` attribute shouldn't be on an element like this if the text content of that element is the same. See https://www.a11yproject.com/posts/2013-04-22-title-attributes/#:~:text=HTML%20title%20attributes%20are%20often,and%20possibly%20not%20even%20used. for reference

src/app/components/DataSelection/DataSelectionFilters.tsx
```diff
@@ -138,12 +148,10 @@ const DataSelectionFilters: FunctionComponent<Props> = ({
                     )
                     .map((option) => (
                       <li className="c-data-selection-available-filters__item">
```
janjaap, line 150: `<li>` Element should have a `key` prop

src/app/components/DataSelection/DataSelectionList/DataSelectionList.tsx
```diff
@@ -0,0 +1,71 @@
...
+            />
+          </Link>
+
+          {row.content.map(
+            (variables, i) =>
+              i !== 0 && (
+                // @ts-ignore
+                <DataSelectionFormatter
+                  // eslint-disable-next-line react/no-array-index-key
+                  key={i}
```
janjaap, line 57: Why not use the value of `variables`?

src/app/components/DataSelection/DataSelectionSbiFilters.tsx
```diff
@@ -113,11 +125,12 @@ const DataSelectionSbiFilters: FunctionComponent<Props> = ({ availableFilters, a
...
                 <li className={`c-sbi-filter__item c-sbi-filter__item--${option.slug}`}>
-                  <button
+                  <StyledLink
+                    inList
                     type="button"
-                    className="c-sbi-filter__button o-btn o-btn--link o-btn--inline-block"
+                    forwardedAs="button"
                     onClick={() => clickFilter(option.id)}
-                    title="{{option.label}}"
+                    title={option.label}
```
janjaap, line 133: `title` Prop should not be on this element

src/app/components/DataSelection/DataSelectionSbiFilters.tsx
```diff
@@ -113,11 +125,12 @@ const DataSelectionSbiFilters: FunctionComponent<Props> = ({ availableFilters, a
               .slice(0, isExpanded ? currentFilter.options.length : showMoreThreshold)
               .map((option) => (
                 <li className={`c-sbi-filter__item c-sbi-filter__item--${option.slug}`}>
```
janjaap, line 127: `<li>` Element should have a `key` prop

src/app/components/SplitScreen/SplitScreen.tsx
```diff
@@ -31,7 +31,7 @@ const SplitScreen: FunctionComponent<SplitScreenProps> = ({
         qa-dashboard__column--right
       `}
     >
-      <div>{rightComponent}</div>
+      <div style={{ width: '100%' }}>{rightComponent}</div>
```
janjaap, line 34: Make it into a styled component. Using object literals as prop values has a negative impact on rendering performance.

src/map/components/detail-result/MapDetailResultWrapper.test.jsx
```diff
@@ -39,10 +39,10 @@ describe('MapDetailResultWrapper', () => {
         </ul>
       </MapDetailResultWrapper>,
     )
-    expect(wrapper).toMatchSnapshot()
+    expect(wrapper.find('map-detail-result__header').length).toBe(0)
```
janjaap, line 42: Shouldn't 'map-detail-result__header' be '.map-detail-result__header'?

src/map/components/detail-result/MapDetailResultWrapper.tsx
```diff
@@ -80,20 +88,15 @@ const MapDetailResultWrapper: FunctionComponent<MapDetailResultWrapperProps> = (
         )}
         {children}
         <footer className="map-search-results__footer">
-          <button
+          <StyledButton
             type="button"
-            className="map-search-results__button"
             onClick={onMaximize}
             title="Volledig weergeven"
```
janjaap, line 94: No `title` attribute

src/map/components/search-results/MapSearchResults.jsx
```diff
@@ -103,20 +112,15 @@ const MapSearchResults = ({
         </ul>
         {!isEmbed && (
           <footer className="map-search-results__footer">
-            <button
+            <StyledButton
               type="button"
-              className="map-search-results__button"
               onClick={onMaximize}
               title="Volledig weergeven"
```
janjaap, line 118: No `title` attribute

src/map/components/shape-summary/ShapeSummary.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const ShapeSummary: FunctionComponent<Props> = ({ shapeDistanceTxt, onClearDrawing }) => (
+  <div className="shape-summary">
+    <span className="shape-summary__label" aria-label={`Lijn: ${shapeDistanceTxt}`}>
+      Lijn:
+      {shapeDistanceTxt}
+    </span>
+
+    <Button
+      title="Lijn verwijderen"
```
janjaap, line 20: Should have an attribute `aria-labelledby` and no `title` attribute

src/map/components/toggle-drawing/ToggleDrawing.jsx
```diff
@@ -10,6 +27,10 @@ const ToggleDrawing = ({
   shapeDistance,
 }) => {
   const expanded = !!(isEnabled || shapeMarkers > 1)
+  const printMode = useSelector(isPrintMode)
+  const embedMode = useSelector(isEmbedded)
+  const embedPreviewMode = useSelector(isEmbedPreview)
```
janjaap, line 32: Using three `useSelector` calls will update and rerender the component three times. Can't these be combined into a single call?

src/map/containers/panel/_map-panel.scss
```diff
@@ -204,16 +195,15 @@
...
-    color: $secondary-gray60;
-    height: 24px !important;
-    line-height: 22px !important;
-    padding: 4px;
-    width: 24px !important;
+    color: $black;
+    height: 40px !important;
+    line-height: 36px !important;
+    width: 40px !important;
+    //font-size: 32px !important;
```
janjaap, line 206: Remove?

src/shared/assets/icons/first.svg
```diff
@@ -1,13 +1,9 @@
...
-    <!-- Generator: Sketch 40.3 (33839) - http://www.bohemiancoding.com/sketch -->
-    <title>icoon-eerste</title>
-    <desc>Created with Sketch.</desc>
-    <defs></defs>
     <g id="Styleguide" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
         <g id="icoon-eerste">
-            <polygon id="Fill-2" var:fill.fill="#fff" points="13.6797546 10.3187297 11 12.9994948 13.6797546 15.6812703 17.9974738 20 20.3700469 17.6264165 16.0523277 13.3097077 15.7421148 12.9994948 16.0523277 12.6892818 20.3700469 8.37257308 17.9974738 6"></polygon>
-            <polygon id="Fill-3" var:fill.fill="#fff" points="6 20 9.00954085 20 9.00954085 6 6 6"></polygon>
+            <polygon id="Fill-2" points="13.6797546 10.3187297 11 12.9994948 13.6797546 15.6812703 17.9974738 20 20.3700469 17.6264165 16.0523277 13.3097077 15.7421148 12.9994948 16.0523277 12.6892818 20.3700469 8.37257308 17.9974738 6"></polygon>
+            <polygon id="Fill-3" points="6 20 9.00954085 20 9.00954085 6 6 6"></polygon>
```
janjaap, line 6: `id` Attributes should not be present in `<svg />`, unless explicitly required. Having more than one svg element on a page can lead to duplicate ids, which is invalid HTML. See, for instance, src/shared/assets/icons/last.svg.

src/shared/assets/icons/maximize.svg
```diff
@@ -1,19 +1,15 @@
...
-            <use id="Fill-1" stroke="#CCCCCC" mask="url(#mask-2)" stroke-width="2" fill="#FFFFFF" xlink:href="#path-1"></use>
-            <polygon id="Fill-2" fill="#666666" points="6.5581535 18.1277262 6.5581535 14.9044828 4.61454979 14.9044828 4.61454979 21.3741077 11.0841748 21.3741077 11.0841748 19.429498 8.00479021 19.429498 10.631472 16.765594 9.25626175 15.3893777"></polygon>
-            <polygon id="Fill-3" fill="#666666" points="14.9043822 4.61495219 14.9043822 6.5585559 17.9837667 6.5585559 15.3560789 9.22346595 16.7312892 10.5986762 19.4293974 7.86032774 19.4293974 11.0845772 21.3730011 11.0845772 21.3730011 4.61495219"></polygon>
+            <polygon id="Fill-2" fill="#000000" points="6.5581535 18.1277262 6.5581535 14.9044828 4.61454979 14.9044828 4.61454979 21.3741077 11.0841748 21.3741077 11.0841748 19.429498 8.00479021 19.429498 10.631472 16.765594 9.25626175 15.3893777"></polygon>
+            <polygon id="Fill-3" fill="#000000" points="14.9043822 4.61495219 14.9043822 6.5585559 17.9837667 6.5585559 15.3560789 9.22346595 16.7312892 10.5986762 19.4293974 7.86032774 19.4293974 11.0845772 21.3730011 11.0845772 21.3730011 4.61495219"></polygon>
```
janjaap, line 12: Same applies here; duplicate id values

janjaap (changes requested):

src/map/components/toggle-drawing/ToggleDrawing.jsx
chrisvanmook: Tested this and it's just being rendered once. How do you think this would cause multiple re-renders? Using multiple hooks doesn't re-render a component, unless the hook has side-effects and updates a state.

src/app/components/ActiveFilters/ActiveFilters.tsx
chrisvanmook: The text content isn't the same here: I've added `Filter verwijderen`. I shall change it to `aria-label`

src/map/components/shape-summary/ShapeSummary.tsx
chrisvanmook: Don't you mean `aria-label`?

janjaap: No `aria-labelledby`. The value should be an id of a label that contains text that describes the purpose of the button. The label can, or probably should, be hidden from non-screen readers

src/map/components/toggle-drawing/ToggleDrawing.jsx
janjaap: When the component renders first, a call to `useSelector` returns the default selector value. When the store has updated, the selector is called again and then returns the value it should return. The component is then rendered again. So, for three `useSelector` calls, there will be multiple rerenders. Maybe it's not a problem for this particular component, but it can be an expensive operation.

janjaap (approved):

src/map/components/toggle-drawing/ToggleDrawing.tsx
```diff
@@ -0,0 +1,90 @@
...
+      clickEvent = onReset
+    }
+  }
+
+  if (printMode || embedMode || embedPreviewMode) {
+    return null
+  }
+
+  return (
+    <>
```
jonkoops, line 63: I think we can drop this fragment, the value returned is a single element.

jonkoops (changes requested):

src/app/components/DataSelection/DataSelectionList/DataSelectionList.tsx
```diff
@@ -0,0 +1,70 @@
...
+    ]
+    formatters: string[]
+  }
+}
+
+const StyledListItem = styled.li`
+  margin-bottom: ${themeSpacing(2)};
+`
+
+export const toDetailPage = (endpoint: string): LocationDescriptor => {
```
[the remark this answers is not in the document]
chrisvanmook: Will have a look!

src/environment.ts
```diff
@@ -3,7 +3,7 @@ const environment = {
   IIIF_ROOT: 'https://acc.images.data.amsterdam.nl/',
   API_ROOT: 'https://acc.api.data.amsterdam.nl/',
   CMS_ROOT: 'https://acc.cms.data.amsterdam.nl/',
-  GRAPHQL_ENDPOINT: 'https://acc.api.data.amsterdam.nl/cms_search/graphql/',
+  GRAPHQL_ENDPOINT: 'http://localhost:8080/cms_search/graphql/',
```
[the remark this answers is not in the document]
chrisvanmook: Whoop yeah no...

src/map/components/toggle-drawing/ToggleDrawing.jsx
jonkoops: I thought that React batches these changes so it doesn't run it unnecessarily.

src/app/components/DataSelection/DataSelectionList/DataSelectionList.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const { type, subtype, id } = getDetailPageData(endpoint)
+  return {
+    pathname: generatePath(routing.dataDetail.path, { type, subtype, id }),
+  }
+}
+
+const DataSelectionList: FunctionComponent<Props> = ({ content }) => {
+  return (
```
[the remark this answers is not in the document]
janjaap: There is an eslint rule for that, you know.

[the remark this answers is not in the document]
jonkoops: > There is an eslint rule for that, you know.

I feel like we should perhaps create a ticket to set up the linting correctly? 😂

jonkoops (approved):

PR #3855, opened 2021-01-15
[DI-124] Auth tests

janjaap:
This PR completely rewrites the tests for the Auth service, making use of the `jest-localstorage-mock` dep and increasing the coverage to 100%.
In addition, some test config files have been turned into TS files.


src/shared/services/auth/auth.ts
```diff
@@ -146,15 +148,20 @@ function catchError() {
...
+  if (!params) {
     return null
   }
 
   const stateToken = sessionStorage.getItem(STATE_TOKEN)
 
   // The state param must be exactly the same as the state token we
   // have saved in the session (to prevent CSRF)
-  const stateTokenValid = params.state && params.state === stateToken
+  const stateTokenValid = Boolean(params.state && params.state === stateToken)
```
jonkoops, line 164: I think we can actually clean this up with some optional chaining:
```suggestion
  const stateTokenValid = params?.state === stateToken
```

src/shared/services/auth/auth.ts
```diff
@@ -96,7 +96,6 @@ export const STATE_TOKEN = 'stateToken'
 export const ACCESS_TOKEN = 'accessToken'
 
 let returnPath = ''
-let tokenData: ParsedToken | null = null
```
jonkoops, line 99: Why are we removing the parsed token? It looks like this would make calls like `getScopes` and `getName` slower as they have to decode the token again.

jonkoops (approved):

src/shared/services/auth/auth.ts
janjaap: Removing it makes testing a whole lot easier; tests were far from transparent and difficult to understand or debug. Trying to get a performance gain from not parsing a JWT is a matter of milliseconds. There are bigger performance issues in the application than that.

jonkoops: For sure, just wanted to know the motivation.

test/setup-jest.ts
```diff
@@ -0,0 +1,22 @@
...
+import 'leaflet'
+import 'leaflet-draw'
+import 'raf/polyfill'
+import 'jest-localstorage-mock'
+import 'isomorphic-fetch'
+
+// React 17 Enzyme adapter
+Enzyme.configure({ adapter: new Adapter() })
+
+declare global {
```
chrisvanmook, line 12: Perhaps better to move this to `setup-env.ts`?

chrisvanmook (approved):

PR #3863, opened 2021-01-18
Refactored detail-panels-visitor spec to use fixtures

chrisvanmook:


jonkoops (approved):

cypress/support/api.ts
```diff
@@ -0,0 +1,141 @@
...
+    ['GET', constructApiURLRegex(`${value.path}([a-z0-9])`), value.singleFixture, anyAlias],
+    ['GET', constructApiURLRegex(`${value.path}(.*)`), value.listFixture, manyAlias],
+  ]
+  return {
+    ...acc,
+    [key]: {
+      intercepts,
+      single: `@${singleAlias}`,
+      any: `@${anyAlias}`,
+      many: value.listFixture ? `@${manyAlias}` : undefined,
```
janjaap, line 127: The API folders use exports `singleFixture` and `listFixture`. `apiFixtures` Exposes `single` and `many`. I would use ubiquitous names here and change `many` to `list`.

Having both `single` and `any` implies that two different fixtures exist where, in fact, they return the same fixture. Do we need both `any` and `single` and, if so, can you add documentation as to what their intended purpose is?

cypress/support/api.ts
```diff
@@ -0,0 +1,141 @@
+import {
```
janjaap, line 1: Wouldn't it be better to have this file (`api.ts`) be in the `src/api` folder? The contents of this file should be in sync with the folder structure of `src/api` (which, in itself is questionable, because it's more work to maintain) and can potentially be used by both unit- and e2e-tests.

cypress/fixtures/kadastraalObjectVisitor.json
```diff
@@ -38,7 +38,7 @@
       "items": ["ASD05 G 08369 A 0001", "ASD05 G 08369 A 0002"]
     },
     {"subheading": "Adressen",
-      "items": ["Nes 100B", "Rokin 87A"]
+      "items": ["Eerste Anjeliersdwarsstraat 1-H"]
```
janjaap, line 41: Why do the `cypress/fixtures` files need to be updated? Can't they be removed, since we're using the data fixtures from `src/api` folder?

cypress/support/api.ts
chrisvanmook: Good point, and will document when `any` can be used

cypress/support/api.ts
chrisvanmook: This file is explicit for cypress now (`intercepts`, and selectors `single`, `any`, `many`). Perhaps I could create `index.ts` inside `./src/api/` that will export [this object](https://github.com/Amsterdam/atlas/pull/3863/files#diff-5b22e0b37aedd68823e03250036c3ea4cbc203173e38d1c76a93fcd349e292ddR29), that can be used in `api.ts` and anywhere inside `src`?

cypress/fixtures/kadastraalObjectVisitor.json
chrisvanmook: Since we have a lot of normalisations and side effects inside our detail pages, we do need to check if the results are normalised properly. These aren't "fixtures" though, just jsons to check the right results. Shall I rename the `fixtures` directory?

janjaap: Well, the point is that we now need both a data fixture as well as a file that contains transformed fixture data. Which means that, when the fixture's data changes, the transformed data also has to be changed. IMHO, the e2e shouldn't verify that specific content is visible, only certain fields. The unit tests should verify that data is correctly tranformed.

cypress/support/api.ts
janjaap: Can't we have a script that reads the `src/api` folder contents and creates an object from that?

cypress/fixtures/kadastraalObjectVisitor.json
chrisvanmook: as discussed, we will leave it for now until we have proper unit tests

scripts/generate-api-index.ts
```diff
@@ -0,0 +1,117 @@
...
+
+function camelize(str?: string) {
+  return str
+    ? str
+        .replace(/(?:^\w|[A-Z]|\b\w)/g, (word, index) => {
+          return index === 0 ? word.toLowerCase() : word.toUpperCase()
+        })
+        .replace(/\s+/g, '')
+    : null
+}
```
jonkoops: I think this would be more readable without a tenary:

```suggestion
function camelize(str?: string) {
  if (!str) {
    return null
  }

  return str
    .replace(/(?:^\w|[A-Z]|\b\w)/g, (word, index) => index === 0 ? word.toLowerCase() : word.toUpperCase())
    .replace(/\s+/g, '')
}
```

jonkoops (approved):

janjaap (approved):

PR #3873, opened 2021-01-20
[DI2-102] Mobile header link fix

janjaap:
This PR fixes an issue on smaller screens where the search overlay would overlap the links in the site header, making them unclickable. The solution was to keep track of the overlay and to apply a specific CSS rule for that overlay so that, only when visible, should span the entire width of the viewport.

To ensure the functionality, integration tests have been added to the homepage spec.


src/app/components/SearchBar/SearchBar.tsx
```diff
@@ -20,7 +20,13 @@ const Z_INDEX_OFFSET = 2 // Set a custom offset
 const StyledSearchBarToggle = styled(SearchBarToggle)`
```
jonkoops, line 20: You should be  able to add the type information here so you don't have to redefine it below.
```suggestion
const StyledSearchBarToggle = styled(SearchBarToggle)<{ searchIsVisible: boolean }>`
```

src/app/components/SearchBar/SearchBar.tsx
```diff
@@ -20,7 +20,13 @@ const Z_INDEX_OFFSET = 2 // Set a custom offset
...
   z-index: 23;
   position: absolute;
-  left: 0;
+
+  ${({ searchIsVisible }: { searchIsVisible: boolean }) =>
```
jonkoops, line 24: So this could be removed
```suggestion
  ${({ searchIsVisible }) =>
```

src/app/components/SearchBar/SearchBar.tsx
```diff
@@ -91,6 +98,10 @@ const SearchBar: FunctionComponent<SearchBarProps> = ({
     placeholder: placeHolder,
   }
 
+  const toggleWidth = (isOpen: boolean): void => {
```
jonkoops, line 101: No need to add `: void` here as the type can be inferred.`
```suggestion
  const toggleWidth = (isOpen: boolean) => {
```

src/header/components/HeaderSearch.tsx
```diff
@@ -86,34 +86,37 @@ const HeaderSearch: FunctionComponent = () => {
...
-    }
 
-    // @ts-ignore
-    document.activeElement?.blur()
-  }
+      // @ts-ignore
+      document.activeElement?.blur()
+    },
+    [inputValue, selectedElement, searchBarFilterValue],
+  )
```
jonkoops, line 119: I believe you can get rid of this `@ts-ignore`:

```suggestion
   if (document.activeElement instanceof HTMLElement) {
     document.activeElement.blur()
   }
```

src/header/components/HeaderSearch.tsx
```diff
@@ -86,34 +86,37 @@ const HeaderSearch: FunctionComponent = () => {
...
+      } else {
+        const queryString = {
+          term: inputValue.trim(),
+        }
+        const query = new URLSearchParams(
+          Object.entries(queryString).reduce(
+            (acc, [key, value]) => (value ? { ...acc, [key]: value } : acc),
+            {},
+          ),
+        )
```
jonkoops, line 106: Can't we put this hash directly into the constructor?
```suggestion
        const query = new URLSearchParams({
          term: inputValue.trim(),
        })
```

jonkoops (changes requested):

src/header/components/HeaderSearch.tsx
janjaap: Possibly, but that code was already in the component.

jonkoops: Sure, but no harm leaving things better than they were before, no?

janjaap: Very true. But making a change like that requires adding tests, which there are none for this component. I'll create a different PR for that.

jonkoops (approved):

src/header/components/HeaderSearch.tsx
jonkoops: I've simplified the code and tested it locally, looks like it was just redundant.

chrisvanmook (approved):

PR #3896, opened 2021-01-25
[DI2-161] Show Alert when trying to get panorama info if API returns a 403

chrisvanmook:


src/app/pages/MapPage/map-search/PanoramaPreview.tsx
```diff
@@ -97,6 +99,9 @@ const PanoramaPreview: FunctionComponent<PanoramaPreviewProps> = ({
   }
 
   if (result.status === PromiseStatus.Rejected) {
+    if (result.error instanceof ForbiddenError) {
```
janjaap, line 102: Can you add coverage for this condition?

src/app/sentry.ts
```diff
@@ -8,5 +8,6 @@ if (process.env.NODE_ENV === 'production') {
   init({
     dsn: 'https://43045d79c42e4eb7a9bdf8e22fff0d9b@sentry.data.amsterdam.nl/29',
     environment: deployEnv,
+    ignoreErrors: [new RegExp(`401 Unauthorized|403 Forbidden`, 'mi')],
```
janjaap, line 11: Did you test this regex?

src/app/utils/useCheckPanoramaPermission.ts
```diff
@@ -0,0 +1,36 @@
...
+  }, [])
+
+  return {
+    showComponent,
+    isForbidden,
+    hasGenericError,
+  }
+}
+
+export default useCheckPanoramaPermission
```
janjaap, line 36: Can you add some test cases for this hook?

src/app/utils/useCheckPanoramaPermission.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const useCheckPanoramaPermission = () => {
+  const [showComponent, setShowComponent] = useState(false)
+  const [isForbidden, setIsForbidden] = useState(false)
+  const [hasGenericError, setHasGenericError] = useState(false)
+
+  useEffect(() => {
+    ;(async () => {
+      try {
+        await fetchWithToken(
```
janjaap, line 15: You can use `fetchProxy` instead; would be best if we get rid of the distinction between fetching with or without token and have one entry point for fetching that appends the bearer token when it's present.

src/app/utils/useCheckPanoramaPermission.ts
```diff
@@ -0,0 +1,36 @@
...
+          `${environment.API_ROOT}${PANORAMA_CONFIG.PANORAMA_ENDPOINT_PREFIX}/TMX7316010203-001187_pano_0000_001517/adjacencies/?newest_in_range=true&tags=mission-bi`,
```
janjaap, line 16: Why are you using this specific URL? This doesn't seem to be a health-check page; it throws errors when it is requested and it weighs over 15KB. Wouldn't it be better to query a known URL that is blocked? Like https://api.data.amsterdam.nl/brk/? The page payload is way smaller than 15KB.

src/app/utils/useCheckPanoramaPermission.ts
```diff
@@ -0,0 +1,36 @@
...
+        )
+        setShowComponent(true)
+      } catch (e) {
```
janjaap, line 19: Why the short variable? Wouldn't `error` or `exception` be more readable?

src/app/utils/useCheckPanoramaPermission.ts
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useState } from 'react'
+import { fetchWithToken } from '../../shared/services/api/api'
+import environment from '../../environment'
+import { PANORAMA_CONFIG } from '../../panorama/services/panorama-api/panorama-api'
+import { ForbiddenError } from '../../shared/services/api/customError'
+
+const useCheckPanoramaPermission = () => {
+  const [showComponent, setShowComponent] = useState(false)
+  const [isForbidden, setIsForbidden] = useState(false)
+  const [hasGenericError, setHasGenericError] = useState(false)
```
janjaap, line 10: Shouldn't these two state values be `undefined` at first? Checking for `false` might lead to false positives when rendering components. Take `isForbidden`; because it is `false` at first, a component might be rendered. When the request below fails, the value might be set to `true` and the components that was rendered initially, will be taken out of the DOM, because of the `403` response.

src/app/utils/useCheckPanoramaPermission.ts
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useState } from 'react'
+import { fetchWithToken } from '../../shared/services/api/api'
+import environment from '../../environment'
+import { PANORAMA_CONFIG } from '../../panorama/services/panorama-api/panorama-api'
+import { ForbiddenError } from '../../shared/services/api/customError'
+
+const useCheckPanoramaPermission = () => {
```
janjaap, line 7: Why not request the permission status when the website loads? That way, we wouldn't have to request the status  each time this hook is used.

src/panorama/sagas/panorama.js
```diff
@@ -59,7 +60,9 @@ export function* handlePanoramaRequest(fn, input, tags) {
     }
     yield put(fetchPanoramaSuccess({ ...panoramaData, tags }))
   } catch (error) {
-    yield put(fetchPanoramaError(error))
+    if (!(error instanceof ForbiddenError)) {
+      yield put(fetchPanoramaError(error))
```
janjaap, line 64: Can you update the test case for this change?

src/shared/services/api/api.ts
```diff
@@ -23,6 +23,10 @@ const handleErrors = (response: Response, reloadOnUnauthorized = false) => {
     throw new AuthError(response.status, '')
   }
 
+  if (response.status === 403) {
+    throw new ForbiddenError(response.status, '')
```
janjaap, line 27: Update the tests, maybe?

janjaap (changes requested):

src/app/components/LocationSearch/LocationSearch.tsx
```diff
@@ -81,20 +80,16 @@ const LocationSearch: FunctionComponent = () => {
...
-                  lng: location.longitude,
-                }}
-                width={438}
-                radius={180}
-              />
-            ) : (
-              <PanoAlert />
-            )
-          ) : null}
+          {numberOfResults && (
```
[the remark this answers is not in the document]
chrisvanmook: As this change is simply a revert from a change a couple of weeks ago, I would leave this out of scope.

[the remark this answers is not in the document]
janjaap: Never hurts to add a couple of test cases 🤷🏻‍♂️

src/app/utils/useCheckPanoramaPermission.ts
```diff
@@ -0,0 +1,36 @@
...
+  const [showComponent, setShowComponent] = useState(false)
+  const [isForbidden, setIsForbidden] = useState(false)
+  const [hasGenericError, setHasGenericError] = useState(false)
+
+  useEffect(() => {
```
jonkoops, line 12: I am quite confident you could replace this with our `usePromise` hook to get the error value.

src/app/utils/useCheckPanoramaPermission.ts
jonkoops: Seem fine to me to keep this `false`, but it seems a bit weird to be requesting a single pano as a way to check if it's possible to see them.

src/panorama/containers/PanoramaWrapper.tsx
```diff
@@ -0,0 +1,36 @@
...
+import PanoramaContainer from './PanoramaContainer'
+import PanoAlert from '../../app/components/PanoAlert/PanoAlert'
+import { ERROR_MESSAGES, ErrorType } from '../../shared/ducks/error/error-message'
+import useCheckPanoramaPermission from '../../app/utils/useCheckPanoramaPermission'
+
+/**
+ * Wrapper to check if user is allowed to view the panorama pictures by requesting
+ * random panorama info.
+ * If the API returns a 403 Forbidden status code, show PanoAlert, otherwise show a generic error
+ * @constructor
```
jonkoops, line 12: This is not a constructor.

src/shared/services/api/api.ts
```diff
@@ -23,6 +23,10 @@ const handleErrors = (response: Response, reloadOnUnauthorized = false) => {
     throw new AuthError(response.status, '')
   }
 
+  if (response.status === 403) {
```
jonkoops, line 26: Instead of creating a new error for each possible status code, perhaps it would be better to have a generic `HttpErrorResponse` class and check the `code` field instead.

jonkoops (changes requested):

src/app/utils/useCheckPanoramaPermission.ts
chrisvanmook: Good point, will change this 👍

src/app/utils/useCheckPanoramaPermission.ts
chrisvanmook: Could you try to explain how we could accomplish this without moving the logic to the highest level of the application?

janjaap: It should be moved to the highest level, right? We need to be able to check if a session has been authorized before we start serving panorama components.

src/shared/services/api/api.ts
chrisvanmook: I agree, but I'll leave it out of scope for now and let's discuss this with the team

src/app/utils/useCheckPanoramaPermission.ts
chrisvanmook: You are completely right! I will remove this

src/environment.ts
```diff
@@ -1,9 +1,9 @@
 const environment = {
   DEPLOY_ENV: 'acceptance',
-  IIIF_ROOT: 'https://acc.images.data.amsterdam.nl/',
-  API_ROOT: 'https://acc.api.data.amsterdam.nl/',
-  CMS_ROOT: 'https://acc.cms.data.amsterdam.nl/',
-  GRAPHQL_ENDPOINT: 'https://acc.api.data.amsterdam.nl/cms_search/graphql/',
+  IIIF_ROOT: 'https://images.data.amsterdam.nl/',
+  API_ROOT: 'https://api.data.amsterdam.nl/',
+  CMS_ROOT: 'https://cms.data.amsterdam.nl/',
+  GRAPHQL_ENDPOINT: 'https://api.data.amsterdam.nl/cms_search/graphql/',
```
jonkoops: These changes should be removed.

src/map/components/detail-result/MapDetailResultWrapper.tsx
```diff
@@ -36,10 +39,15 @@ const MapDetailResultWrapper: FunctionComponent<MapDetailResultWrapperProps> = (
   const { panoramaUrl, link, linkComponent } = useGetLegacyPanoramaPreview(location as any)
   const isEmbed = useSelector(isEmbedded)
   const user = useSelector(getUser)
+  // const { isForbidden, showComponent } = useCheckPanoramaPermission()
 
+  const result = usePromise(
+    () => fetchProxy('https://acc.api.data.amsterdam.nl/brk/?format=json'),
```
jonkoops, line 52: Perhaps a comment explaining why we're making a call to acceptance APIs would be helpful here.

src/map/components/detail-result/MapDetailResultWrapper.tsx
```diff
@@ -36,10 +39,15 @@ const MapDetailResultWrapper: FunctionComponent<MapDetailResultWrapperProps> = (
   const { panoramaUrl, link, linkComponent } = useGetLegacyPanoramaPreview(location as any)
   const isEmbed = useSelector(isEmbedded)
   const user = useSelector(getUser)
+  // const { isForbidden, showComponent } = useCheckPanoramaPermission()
```
jonkoops: Can we get rid of this comment?

src/map/components/detail-result/MapDetailResultWrapper.tsx
```diff
@@ -1,19 +1,22 @@
...
+import usePromise from '../../../app/utils/usePromise'
+import { ForbiddenError } from '../../../shared/services/api/customError'
 
 export interface MapDetailResultWrapperProps {
   title: string
   location?: LatLngLiteral
   subTitle?: string | null
   onMaximize: () => void
-  children: React.ReactNode
+  children: ReactNode
```
jonkoops, line 20: This can be removed entirely as the `FunctionComponent` type automatically adds a `children` prop.

src/map/components/search-results/MapSearchResults.jsx
```diff
@@ -39,10 +42,14 @@ const MapSearchResults = ({
     }))
 
   const { panoramaUrl, link, linkComponent } = useGetLegacyPanoramaPreview(location)
+  const result = usePromise(
```
jonkoops, line 54: Same here, might make sense to use a comment. Perhaps we should but this invocation in a separate file so it can be reused.

tsconfig.json
```diff
@@ -8,6 +8,9 @@
   ],
   "compilerOptions": {
     "strict": true,
+    "target": "ES2020",
+    "module": "es2020",
```
jonkoops, line 12: Pet peeve
```suggestion
    "module": "ES2020",
```

jonkoops (changes requested):

src/app/utils/useCheckPanoramaPermission.ts
jonkoops: IMHO this is fine for now as a temporary workaround, it doesn't have a major impact on performance or anything.

chrisvanmook: I do not fully agree with you:
1. We will have to move logic coupled to Panorama's to a higher level: we would introduce a side effect for the whole application. 
2. Users can connect to VPN after they've opened the site.
3. We will force ALL users to make this request, so performance wise this would also not be ideal.
4. We really won't win much performance: the request is ~700 bytes and will only be done on first render, not blocking other components and is used on components (PanoramaPreview) which aren't super essential  

I would say let's keep it the way it works now. This is a temporary fix and IMHO we should start focus soon on other performance wins

jonkoops (approved):

PR #3900, opened 2021-01-27
[DI2-126] Homepage tests

janjaap:
This PR fixes the homepage e2e tests by
- adding `data-testid` attributes where necessary
- changing the `checkMenuLink` function parameters for readability
- rewriting the way footer links were checked


cypress/integration/drawing_spec.ts
```diff
@@ -53,6 +55,9 @@ describe('Drawing', () => {
       cy.get(DRAWING.buttonRemove).click()
       cy.get(DRAWING.polygon).should('have.length', 1).and('be.visible')
       cy.get(DRAWING.buttonCloseDrawTool).click()
+
+      cy.pause()
```
chrisvanmook, line 59: why do we need the pause here?

src/app/components/HomePage/services/navigationLinks.tsx
```diff
@@ -39,6 +40,7 @@ const navigationLinks: NavigationLink[] = [
         <Map />
       </Icon>
     ),
+    testId: 'Kaart',
```
chrisvanmook, line 43: I'm not sure if there's a naming convention on `data-testid`, but now we mostly use camelCase, while these are PascalCase. I suggest to choose just one, and maybe use camelCase for id's. Also, everything in code is written in english so you might just want to have the value `map` here

janjaap: The `testId` values in this configuration object are prefixed in the `HeaderMenu` component, that's why the test ids are capitalized here.

chrisvanmook (approved):

PR #3901, opened 2021-01-28
[DI2-126] Map tests

janjaap:
This PR fixes the failing map integration tests. Note that the Cypress method `checkAllLayers` has been removed; this function verified the working of a specific API endpoint, which is something the integration tests shouldn't be responsible for.


cypress/integration/map_spec.ts
```diff
@@ -122,7 +123,8 @@ describe('map module', () => {
       cy.get(MAP.mapOverlayPane).children().should('exist')
       cy.get(MAP.mapOverlayPane).find('canvas').should('exist')
     })
-    it('should add a layer to the map', () => {
+
+    it.only('should add a layer to the map', () => {
```
chrisvanmook, line 127: Remove `.only`?

cypress/support/map-commands.ts
```diff
@@ -52,104 +51,29 @@ Cypress.Commands.add('checkTopography', () => {
...
+  cy.get(`[data-testid="mapLegendLayerButton${categoryTestId || category}"]`)
     .contains(category)
     .parents(MAP.mapLegendItemButton)
     .click('right')
+
   cy.get(MAP.mapZoomIn).click()
   cy.get(MAP.imageLayer).should('not.exist')
 })
 
-Cypress.Commands.add(
```
chrisvanmook, line 67: 🎉

chrisvanmook (approved):

PR #3905, opened 2021-01-29
[DI2-126] Panorama spec

janjaap:
This PR fixes a couple of issues in the panorama e2e spec. Note that one test case has been skipped, because of unclarity of what has to be verified with that test.


src/panorama/components/PanoramaToggle/PanoramaToggle.jsx
```diff
@@ -22,13 +22,13 @@ const PanoramaToggle = ({
     showMenuToggle(null)
   }, [currentLabel])
 
-  const handleOpenPanoramaExternal = () => {
+  const handleOpenPanoramaExternal = useCallback(() => {
```
jonkoops, line 25: Why do we need to wrap this method in a `useCallback`? We already discussed the performance in other PRs right? Also it's used in an anonymous method further down.

jonkoops (changes requested):

src/panorama/components/PanoramaToggle/PanoramaToggle.jsx
janjaap: I guess we don't need it here.

cypress.json
```diff
@@ -1,8 +1,8 @@
 {
   "baseUrl": "http://localhost:3000",
-  "defaultCommandTimeout": 10000,
+  "defaultCommandTimeout": 3000,
   "projectId": "k3sms4",
-  "requestTimeout": 15000,
+  "requestTimeout": 3000,
```
chrisvanmook, line 5: Where are these based on? Can't we just remove then and use the default values?

cypress/integration/panorama_spec.ts
```diff
@@ -90,16 +80,16 @@ describe('panorama module', () => {
...
-      cy.get(PANORAMA.panoramaMenu).should('not.exist')
       cy.get(PANORAMA.panoramaToggle).first().click()
       cy.get(PANORAMA.panoramaMenu).find('[aria-hidden="false"]').should('exist')
       cy.get(PANORAMA.panoramaMenu).find('ul > li').eq(2).click()
       cy.get(PANORAMA.panoramaMenu).find('[aria-hidden="true"]').should('exist')
     })
   })
 
-  describe('user should be able to interact with the panorama', () => {
+  describe.skip('user should be able to interact with the panorama', () => {
```
chrisvanmook, line 92: Perhaps add a Todo with a ticket number?

chrisvanmook (approved):

cypress/integration/panorama_spec.ts
janjaap: There is a ticket on the board; I don't think ticket numbers should be referenced in the code.

cypress.json
janjaap: Based on a very slow network, I guess. I reduced the waiting time, because tests were taking very long to complete. And I think there should be at least some time before the request can be considered failed, so I set it to 3000 🤷🏻

jonkoops (approved):

PR #3920, opened 2021-02-01
Bugfix/editorial typeerror

chrisvanmook:


src/app/components/EditorialResults/EditorialResults.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+import EditorialCard from '../EditorialCard/EditorialCard'
+import { toSpecialDetail } from '../../../store/redux-first-router/actions'
+
+jest.mock('../EditorialCard/EditorialCard', () => {
+  return jest.fn(() => <div data-testid="editorialCard" />)
+})
+
+describe('EditorialResults', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
```
jonkoops, line 17: Doesn't this happen automatically?

jonkoops (approved):

src/app/components/EditorialResults/EditorialResults.test.tsx
chrisvanmook: no it doesn't. i.e. it remembers how many times the component has been called before.

src/app/components/EditorialResults/EditorialResults.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    it('should set the correct props for publications', () => {
+      const props = {
+        ...defaultProps,
+        type: CmsType.Publication,
+        results: [{ ...result, type: CmsType.Publication }],
+      }
+
+      render(withAppContext(<EditorialResults {...props} />))
+
+      expect(EditorialCard).toHaveBeenNthCalledWith(
```
janjaap, line 139: It's not very important to check with which parameters the function component was called; more important is what is rendered. In the case of the `EditorialResults` component, the only thing that logically needs to be tested is how many component are rendered. You could render different `data-testid` attribute values for the different types to verify that an amount of card of a specific type has been rendered.

src/app/components/EditorialResults/EditorialResults.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    it('should set the correct props for specials', () => {
+      const props = {
+        ...defaultProps,
+        type: CmsType.Special,
+        results: [{ ...result, type: CmsType.Special }],
+      }
+
+      render(withAppContext(<EditorialResults {...props} />))
+
+      expect(EditorialCard).toHaveBeenNthCalledWith(
```
janjaap, line 167: Same here

src/app/components/EditorialResults/EditorialResults.tsx
```diff
@@ -49,12 +62,14 @@ const EditorialResults = ({
...
   return (
+    // @ts-ignore
     <EditorialCardContainer className={className}>
       {loading ? (
         <LoadingSpinner />
       ) : (
         <>
           {!hasLoadingError &&
+            results &&
             results.length > 0 &&
```
janjaap, line 73: Or just `results?.length`

src/app/components/EditorialResults/EditorialResults.tsx
```diff
@@ -49,12 +62,14 @@ const EditorialResults = ({
   const unauthorizedLabels = getUnauthorizedLabels(matchingErrors)
 
   return (
+    // @ts-ignore
```
janjaap, line 65: What needs to be ignored?

janjaap:
@chrisvanmook Create a ticket for this bug, please.

src/app/components/EditorialCard/EditorialCard.tsx
```diff
@@ -201,22 +202,21 @@ const MetaText = styled(Paragraph)`
...
   showContentType?: boolean
   highlighted?: boolean
 }
 
+export const EDITORIAL_CARD_TEST_ID = 'editorialCard'
+
 const EditorialCard: FunctionComponent<EditorialCardProps> = ({
   title,
-  description,
+  teaser,
```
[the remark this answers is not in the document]
chrisvanmook: Yes, they do. We only changed this field for no particular reason in the front-end so I've changed this not only to avoid confusion, but also because we could now pick the type from `CmsResultItem` instead of re-defining it.

src/app/components/EditorialCard/EditorialCard.tsx
```diff
@@ -227,11 +227,14 @@ const EditorialCard: FunctionComponent<EditorialCardProps> = ({
   highlighted = false,
   ...otherProps
 }) => {
+  if (!type) {
```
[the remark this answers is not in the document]
chrisvanmook: Could perhaps be `null` apparently, we never know what the API returns so best to play safe.

src/app/components/EditorialResults/EditorialResults.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+import EditorialResults, { IMAGE_SIZE } from './EditorialResults'
+import { CMSResultItem } from '../../utils/useFromCMS'
+import { LOADING_SPINNER_TEST_ID } from '../LoadingSpinner/LoadingSpinner'
+import withAppContext from '../../utils/withAppContext'
+import EditorialCard from '../EditorialCard/EditorialCard'
+import { toSpecialDetail } from '../../../store/redux-first-router/actions'
+
+jest.mock('../EditorialCard/EditorialCard', () => {
+  return jest.fn(() => <div data-testid="editorialCard" />)
+})
```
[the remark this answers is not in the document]
chrisvanmook: Because I don't want to test logic in this component + now I can test what kind of props will be set on this component from `EditorialResults`

src/app/components/EditorialResults/EditorialResults.tsx
chrisvanmook: Apparently that still leaves TS errors...

src/app/components/EditorialResults/EditorialResults.tsx
chrisvanmook: An issue with styled-components causing a false positive TS error: https://github.com/Amsterdam/amsterdam-styled-components/issues/331

src/app/components/EditorialResults/EditorialResults.test.tsx
```diff
@@ -92,94 +102,157 @@ describe('EditorialResults', () => {
...
+          extensions: {
+            label: 'No access',
+            code: 'UNAUTHORIZED',
+          },
+        },
+      ],
+    }
+    const { getByTestId } = render(withAppContext(<EditorialResults {...props} />))
+
+    expect(getByTestId('auth-alert')).toBeDefined()
```
janjaap, line 126: To check if an element is actually rendered, you should be using
```jsx
expect(getByTestId('auth-alert')).toBeInTheDocument()
```
Using `toBeDefined()` can give false-positives, for instance when using `getAllByTestId()`, because it always returns a `NodeList` which is always defined.

src/app/components/EditorialResults/EditorialResults.test.tsx
```diff
@@ -92,94 +102,157 @@ describe('EditorialResults', () => {
...
+        {
+          message: 'Just an error',
+          path: ['articleSearch'],
+        },
+      ],
+    }
+    const { getByTestId } = render(withAppContext(<EditorialResults {...props} />))
+
+    expect(getByTestId(ERROR_MESSAGE_TEST_ID)).toBeDefined()
+    fireEvent.click(getByTestId(ERROR_MESSAGE_RELOAD_BUTTON_TEST_ID))
```
janjaap, line 146: You should verify, if possible, that `window.location.reload` hasn't been called prior to the `click` event

src/app/components/EditorialResults/EditorialResults.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+      loading: false,
+      isOverviewPage: false,
+      type: CmsType.Article,
+      errors: [],
+    }
+    const { queryAllByTestId, rerender } = render(
+      withAppContext(<EditorialResults {...props} results={[]} />),
+    )
+
+    expect(queryAllByTestId('editorialCard')).toEqual([])
```
janjaap, line 76: `toHaveLength(0)`?

src/app/components/EditorialResults/EditorialResults.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+    // Should render two cards
+    rerender(
+      withAppContext(<EditorialResults {...props} results={[result, { ...result, id: '2' }]} />),
+    )
+    expect(queryAllByTestId('editorialCard').length).toBe(2)
```
janjaap, line 82: `toHaveLength(2)`

src/app/components/EditorialResults/EditorialResults.tsx
```diff
@@ -71,16 +86,21 @@ const EditorialResults = ({
                 type,
               } = result
 
+              // Since it's possible that a result might not have a slug, we have to skip that one
+              if (!slug || !type) {
+                return null
+              }
```
janjaap, line 92: You could also filter the `results` array so that you don't need to return `null` explicitly

src/app/components/EditorialResults/EditorialResults.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+        queryByTestId(EDITORIAL_CARD_TEST_ID)?.childNodes[0] as Element,
+      )
+      expect(style.paddingTop).toBe('8px')
+
+      // It should build the right URL
+      expect(queryByTestId(EDITORIAL_CARD_TEST_ID)?.closest('a')).toHaveAttribute(
+        'href',
+        '/dossiers/dossier/slug/1/',
+      )
+    })
```
janjaap, line 257: These tests should be part of the `EditorialCard` test suite; you're testing logic that is specific to the `EditorialCard` component and cannot and should not be known to the `EditorialResults` component. If we decide to change what is rendered inside the `EditorialCard` component, the test cases for the `EditorialResults` also might have to be changed.

janjaap (changes requested):

src/app/components/EditorialResults/EditorialResults.test.tsx
chrisvanmook: `getByTestId` doesn't always return a `NodeList`, `queryBy*` does. When this cannot be found, an error will be thrown and thus the test will fail

janjaap: True, but to make sure the element is in the DOM, you can use `toBeInTheDocument()` instead of `toBeDefined()`.

janjaap (approved):

PR #3923, opened 2021-02-03
(Chore) ADR Elasticsearch

janjaap:
This PR contains the decision description of the use of Elasticsearch for the CMS content.


docs/adr/000-historic-decisions.md
```diff
@@ -0,0 +1,46 @@
...
+
+To also offer background stories and articles related to the data from the municipality, the decision was made to introduce a CMS by which other content could be published on data.amsterdam.nl. The different types of content available<sup>1</sup> are:
+- articles
+- publications
+- specials and
+- collections
+
+#### Typeahead APIs
+
+Several APIs within the municipality of Amsterdam offer searching through their data by full text search. All of that data is indexed by Elasticsearch. It concerns the following APIs:
```
jonkoops, line 23: This data is not indexed by our team, I am not sure if the backend team uses ElasticSearch.

docs/adr/000-historic-decisions.md
```diff
@@ -0,0 +1,46 @@
...
+
+- [BAG](https://api.data.amsterdam.nl/atlas/typeahead/bag/)
+- [HR](https://api.data.amsterdam.nl/handelsregister/typeahead/)
+- [BRK](https://api.data.amsterdam.nl/atlas/typeahead/brk/)
+- [Gebieden](https://api.data.amsterdam.nl/atlas/typeahead/gebieden/)
+- [Meetbouten](https://api.data.amsterdam.nl/meetbouten/typeahead/)
+- [Monumenten](https://api.data.amsterdam.nl/monumenten/typeahead/)
+
+The proxy API [api.data.amsterdam.nl/typeahead](https://api.data.amsterdam.nl/typeahead) returns results for all of the above as well as results from the [Dataset catalogue](https://api.data.amsterdam.nl/dcatd/datasets). The `atlas` application uses an autosuggest component that shows the output of the proxy API.
```
jonkoops, line 32: Perhaps it's good to know that this API calls CMS Search for data.

jonkoops (approved):

docs/adr/000-historic-decisions.md
janjaap: The ADR doesn't mention that it is indexed by our team. It only states that it is indexed. I know for a fact that the other APIs use Elasticsearch. That is the main reason why the CMS content needed to be in ES as well.

docs/adr/000-historic-decisions.md
janjaap: Yes, the section below mentions that.

docs/adr/000-historic-decisions.md
jonkoops: On what are you basing that is the main reason? To my knowledge we are the only consumers of this Elastic index. The fact that there are other backend services that use Elastic (not our index) seems unrelated.

Personally I would not document the stack any further than what we have ownership over as it is prone to change. All we need to know is that there are 'typeahead APIs'.

janjaap: I'm basing that on discussions with responsible parties within the Datapunt organisation. 
It is important to mention that the requirement for Elasticsearch on the CMS part is based on the choice for Elasticsearch by other services within the department. Those services and in particular the `typeahead` API, that we are using in the auto-suggest component, all query Elasticsearch indexes and, as I understood it, the obvious route to take was to expose the CMS content via an ES index.
The situation of the infra structure is important to note, because it indicates the relationship between the different parts. That it can change is not a reason not to document it, because everything can be subject to change.

alimpens (approved):

chrisvanmook (approved):

PR #3925, opened 2021-02-03
Updated panoTags to handle legacy URLs

chrisvanmook:


src/app/pages/MapPage/query-params.test.ts
```diff
@@ -0,0 +1,29 @@
+import { panoTagParam } from './query-params'
+
+describe('Map Query Parameters', () => {
+  describe('Panorama', () => {
+    describe('tags', () => {
```
jonkoops, line 5: Perhaps you could use a single `describe('panoTagParam')` call here?

src/app/pages/MapPage/query-params.test.ts
```diff
@@ -0,0 +1,29 @@
...
+      it('should encode the value to parameter value', () => {
```
jonkoops, line 6: Can we agree to use verbs here instead of 'should'? I would prefer this for example:
```suggestion
      it('encodes the value to parameter value', () => {
```

What do you guys think?

janjaap (approved):

janjaap:
@chrisvanmook Unrelated test is failing

jonkoops (changes requested):

jonkoops (approved):

src/app/pages/MapPage/query-params.test.ts
janjaap: Agree

janjaap: `should` Implies may or may not.

PR #3927, opened 2021-02-04
Added detail page for "woningbouwplannen/gebied_bouwblok_woningen"

chrisvanmook:


src/map/services/map-services.config.ts
```diff
@@ -2364,6 +2365,44 @@ const servicesByEndpointType: { [type: string]: ServiceDefinition } = {
       ],
     }),
   },
+  [endpointTypes.woningbouwplannenGebiedBouwblokWoningen]: {
+    type: 'woningbouwplannen/gebied_bouwblok_woningen',
+    endpoint: 'v1/woningbouwplannen/gebied_bouwblok_woningen',
```
janjaap, line 2370: I think the URL should end with a `/`.

chrisvanmook: All the endpoints in this "config" file don't have a trailing `/`. In the logic of the detail page, an `id` will be added to this endpoint. I've created a ticket to create a plan to cleanup and restructure these files.

janjaap: Kewl

janjaap (approved):

jonkoops (approved):

PR #3941, opened 2021-02-09
[DI2-119] Publication date fix

janjaap:
This PR fixes an issue where an article's publication date wouldn't be shown correctly. The cause of this was that the code wasn't using the value of `field_publication_day` which rendered the date always with day 1.

In order to fix the issue, the tests for `useNormalizedCMSResults.js` have been rewritten and the file has been refactored by moving functionality out of the main function so it can be tested separately. The functionality slightly changed in that passing in a value for `field_publication_year` and none for `field_publication_month` will show the year (instead of throwing an error). Passing in a value for both `field_publication_year` and `field_publication_month` will always show 1-{month}-{year}.

### DIFF
--- a/src/normalizations/cms/useNormalizedCMSResults.js
+++ b/src/normalizations/cms/useNormalizedCMSResults.js
@@ -48,11 +48,13 @@ export const getLocaleFormattedDate = ({
 
   const day = parseInt(field_publication_day, 10)
   const monthIndex = parseInt(field_publication_month, 10)
+  const monthIsValidNumber = Number.isNaN(monthIndex) === false
+  const dayIsValidNumber = Number.isNaN(day) === false
 
   const dateParts = [
     year,
-    !Number.isNaN(monthIndex) && Math.max(0, monthIndex - 1),
-    !Number.isNaN(day) && day,
+    monthIsValidNumber && Math.max(0, monthIndex - 1),
+    dayIsValidNumber && day,
   ].filter(Number.isFinite)
 
   const localeDate = new Date(Date.UTC(...dateParts))
--- a/src/normalizations/cms/useNormalizedCMSResults.js
+++ b/src/normalizations/cms/useNormalizedCMSResults.js
@@ -42,7 +42,7 @@ export const getLocaleFormattedDate = ({
   if (field_publication_date) {
     return {
       localeDate: field_publication_date,
-      localeDateFormatted: formatDate(new Date(field_publication_date.replace(' ', 'T'))),
+      localeDateFormatted: formatDate(new Date(field_publication_date)),
     }
   }
 
--- a/src/normalizations/cms/useNormalizedCMSResults.js
+++ b/src/normalizations/cms/useNormalizedCMSResults.js
@@ -63,6 +63,29 @@ export const getLocaleFormattedDate = ({
   }
 }
 
+export const getLinkProps = ({ type, uuid, field_link, field_special_type, title }, slug) => {
+  let to = {}
+
+  if (EDITORIAL_DETAIL_ACTIONS[type]) {
+    if (type === CmsType.Special) {
+      to = EDITORIAL_DETAIL_ACTIONS[type](uuid, field_special_type, slug)
+    } else {
+      to = EDITORIAL_DETAIL_ACTIONS[type](uuid, slug)
+    }
+  }
+
+  let linkProps = { to, forwardedAs: pickLinkComponent(to) }
+  const externalUrl = field_link?.uri ? cleanupDrupalUri(field_link?.uri) : null
+
+  linkProps = externalUrl ? { href: externalUrl, forwardedAs: 'a' } : linkProps
+  linkProps = { ...linkProps, title } // Add the title attribute by default
+
+  return {
+    linkProps,
+    to,
+  }
+}
+
 export const normalizeObject = (data) => {
   const {
     uuid,
@@ -76,7 +99,6 @@ export const normalizeObject = (data) => {
     field_special_type,
     field_file,
     media_image_url,
-    field_link,
     field_links,
     field_related,
     ...otherFields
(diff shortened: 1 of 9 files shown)


src/normalizations/cms/useNormalizedCMSResults.js
```diff
@@ -20,11 +20,73 @@ export const EDITORIAL_FIELD_TYPE_VALUES = {
...
-  return uri.replace(/&amp;/g, '&')
+const cleanupDrupalUri = (uri) => uri.replace(/&amp;/g, '&')
+
+export const getLocaleFormattedDate = ({
+  field_publication_date,
+  field_publication_day,
+  field_publication_year,
+  field_publication_month,
+} = {}) => {
+  const year = parseInt(field_publication_year, 10)
```
jonkoops, line 31: I've created a ticket to see if we can get this as a number straight from the Drupal API.

src/normalizations/cms/useNormalizedCMSResults.js
```diff
@@ -20,11 +20,73 @@ export const EDITORIAL_FIELD_TYPE_VALUES = {
...
+  const yearIsValidNumber = Number.isNaN(year) === false
```
jonkoops, line 32: Would negating this boolean not be shorter?

```suggestion
  const yearIsValidNumber = !Number.isNaN(year)
```

src/normalizations/cms/useNormalizedCMSResults.js
```diff
@@ -20,11 +20,73 @@ export const EDITORIAL_FIELD_TYPE_VALUES = {
...
+  }
+
+  const day = parseInt(field_publication_day, 10)
+  const monthIndex = parseInt(field_publication_month, 10)
+
+  const dateParts = [
+    year,
+    !Number.isNaN(monthIndex) && monthIndex - 1,
+    !Number.isNaN(day) && day,
+  ].filter(Boolean)
```
jonkoops, line 56: Since months are zero indexed in JavaScript dates would this not lead to a missing month number for January? If `monthIndex` is 1 (so January) and `monthIndex - 1` (so `0`) is passed to the `Boolean` filter will this not result in it getting filtered out?

In this case I think a procedural approach here might be more clear and less error prone:

```suggestion
  const dateParts = [year]
  
  if (!Number.isNaN(monthIndex)) {
    dateParts.push(monthIndex - 1)
  }
  
  // Also check if the month was pushed before, we don't
  // want to accidentally add a day in the month position.
  if (dateParts.length === 2 && !Number.isNaN(day)) {
    dateParts.push(day)
  }
```

src/normalizations/cms/useNormalizedCMSResults.js
```diff
@@ -20,11 +20,73 @@ export const EDITORIAL_FIELD_TYPE_VALUES = {
...
+
+  const localeDate = new Date(Date.UTC(...dateParts))
```
jonkoops, line 58: Why do we need to run this through as a UTC date? Do we end up on the wrong day in local time when formatting this later otherwise?

src/normalizations/cms/useNormalizedCMSResults.test.js
```diff
@@ -1,20 +1,100 @@
...
+      const { localeDate, localeDateFormatted } = getLocaleFormattedDate({
+        field_publication_year: 'foobar',
+      })
+
+      expect(localeDate).toEqual('')
+      expect(localeDateFormatted).toEqual('')
+    })
+
+    it('returns an object with dates from field_publication_date', () => {
+      const field_publication_date = '2020-12-02T16:00:00+01:00'
```
jonkoops, line 43: I noticed that there was the following line to normalize the `field_publication_date` field:

```ts
field_publication_date.replace(' ', 'T')
```

It doesn't look like this string replacement logic is actually used according to the tests. so can it be removed?

jonkoops (changes requested):

src/normalizations/cms/useNormalizedCMSResults.js
janjaap: Definitely shorter, but not necessarily more readable.

src/normalizations/cms/useNormalizedCMSResults.js
janjaap: Good one; I'll add a test case for that.

src/normalizations/cms/useNormalizedCMSResults.test.js
janjaap: I'm not sure why it is in place; it was already in the code. Might be required, might also not be. Must be there for a reason, so I'd rather leave it in.

src/normalizations/cms/useNormalizedCMSResults.js
janjaap: I don't know; was already in the code.

src/normalizations/cms/useNormalizedCMSResults.js
janjaap: Solved it in a different way; a bit cleaner than adding conditions imho

src/normalizations/cms/useNormalizedCMSResults.js
jonkoops: It's being used in other parts of the code like this already though (see the `dateParts` variable).

src/normalizations/cms/useNormalizedCMSResults.js
jonkoops: Fair enough, this is probably fine.

src/normalizations/cms/useNormalizedCMSResults.test.js
jonkoops: If that is the case can we add a test for this scenario as well? Just to cover our asses in case we do some refactoring on this code.

src/normalizations/cms/useNormalizedCMSResults.js
jonkoops: Nice, long as we cover this edge case I am happy :smile:

src/normalizations/cms/useNormalizedCMSResults.js
janjaap: If we only have a single value, for instance `new Date(2021)`, the value is considered to be a timestamp and will return the wrong date. Using `new Date(Date.UTC(2021))` will solve that.

src/normalizations/cms/useNormalizedCMSResults.test.js
janjaap: Tested it and removed the `replace` function; the `Date` constructor is able to deal with either `2019-07-22T11:31:48+00:00` or `2019-07-22 11:31:48+00:00`

src/normalizations/cms/useNormalizedCMSResults.js
jonkoops: Ahhhh, yes that makes sense. I guess we could also use `Date.parse` for that.

jonkoops (approved):

PR #3948, opened 2021-02-10
[DI2-188] Add a new section to the feedback modal to request new features

jonkoops:


src/app/components/Modal/FeedbackModal.jsx
```diff
@@ -83,23 +83,37 @@ const FeedbackModalComponent = ({
     </ModalBlock>
     <Divider gutter />
     <ModalBlock>
-      <Heading as="h4">Probleem of suggestie?</Heading>
+      <Heading as="h4">Verzoek nieuwe data of functionaliteiten</Heading>
+      <Paragraph>
+        Heb je een verzoek voor bijvoorbeeld een nieuwe dataset, kaartlaag of een verbetering van de
+        site? We horen het graag.
+      </Paragraph>
+      <Button
```
alimpens, line 91: Does it make sense that you leave this page when you click this button? The other buttons open your mail program (if you have one), you don't leave the page.

jonkoops: Yeah good point, should I add a blank target here so it opens in a new tab?

alimpens: Makes sense to me, but maybe check with UX

chrisvanmook (approved):

src/app/components/Modal/FeedbackModal.jsx
jonkoops: Checked with Martijn and this is accepted behaviour

PR #3954, opened 2021-02-10
Fix issue where loading state for CMS content was set incorrectly

jonkoops:


src/app/utils/useFromCMS.ts
```diff
@@ -89,13 +89,15 @@ function useFromCMS<T = CMSResultItem[]>(
...
           result = normalizeCMSResults(result)
         }
+
         setResults(result)
+        setLoading(false)
       })
       .catch(() => {
         setError(true)
+        setLoading(false)
       })
```
chrisvanmook, line 99: could we use `.finally` here?

src/app/utils/useFromCMS.ts
```diff
@@ -89,13 +89,15 @@ function useFromCMS<T = CMSResultItem[]>(
         } else {
           result = normalizeCMSResults(result)
         }
+
         setResults(result)
+        setLoading(false)
       })
       .catch(() => {
         setError(true)
+        setLoading(false)
```
janjaap, line 98: Could be in a `finally` clause, right?

chrisvanmook (approved):

PR #3971, opened 2021-02-11
Changed parameters for panorama viewer

chrisvanmook:


src/app/components/PanoramaViewer/PanoramaViewer.tsx
```diff
@@ -65,26 +69,28 @@ const PanoramaViewer: FunctionComponent = () => {
...
+    [panoPitch, panoHeading, setPanoImageDate, marzipanoViewer, onClickHotspot],
   )
 
   // Update the URL queries when the view changes
   useEffect(() => {
     if (currentMarzipanoView) {
-      setPano(currentMarzipanoView, 'replace')
+      setPanoHeading(currentMarzipanoView.heading, 'replace')
+      setPanoPitch(currentMarzipanoView.pitch, 'replace')
+      setPanoFov(currentMarzipanoView.fov, 'replace')
```
janjaap, line 93: `fov` Wasn't used, was it?

src/app/pages/MapPage/MapPage.tsx
```diff
@@ -60,18 +66,20 @@ const MapPage: FunctionComponent = () => {
...
   const [center, setCenter] = useParam(centerParam)
   const [zoom, setZoom] = useParam(zoomParam)
-  const [pano] = useParam(panoParam)
+  const [location] = useParam(locationParam)
+  const [panoPitch] = useParam(panoPitchParam)
+  const [panoHeading] = useParam(panoHeadingParam)
   // TODO: Import 'useMatchMedia' directly once this issue has been resolved: https://github.com/Amsterdam/amsterdam-styled-components/issues/1120
   const [showDesktopVariant] = hooks.useMatchMedia({ minBreakpoint: 'tabletM' })
-  const panoActive = !!pano
+  const panoActive = panoHeading !== null && location !== null
```
janjaap, line 74: Params are strings, can they be `null` then?

src/app/pages/MapPage/map-search/PanoramaPreview.tsx
```diff
@@ -113,14 +119,9 @@ const PanoramaPreview: FunctionComponent<PanoramaPreviewProps> = ({
...
-            panoParam,
-            {
-              heading: result?.value?.heading || 0,
-              pitch: 0,
-              fov: PANORAMA_CONFIG.DEFAULT_FOV,
-            },
-          ],
+          [panoPitchParam, panoPitchParam.initialValue],
+          [panoFovParam, panoFovParam.initialValue],
+          [panoHeadingParam, result?.value?.heading],
```
janjaap, line 124: `panoHeadingParam` Shouldn't default to `0`?

src/app/pages/MapPage/query-params.ts
```diff
@@ -104,11 +104,27 @@ export const baseLayerParam: UrlParam<BaseLayer> = {
...
 
-export const panoParam: UrlParam<Pano | null> = {
-  name: 'pano',
+export const panoPitchParam: UrlParam<number | null> = {
+  name: 'pitch',
   defaultValue: null,
-  decode: (value) => value && JSON.parse(value),
-  encode: (value) => value && JSON.stringify(value),
+  initialValue: 10,
+  decode: (value) => parseFloat(value) ?? null,
```
janjaap, line 111: `parseFloat` can return `NaN` which is neither `undefined` or `null`. This will make the `decode` function return `NaN`. Is that intended?

src/app/pages/MapPage/query-params.ts
```diff
@@ -104,11 +104,27 @@ export const baseLayerParam: UrlParam<BaseLayer> = {
...
+  encode: (value) => (typeof value === 'number' ? value.toString() : null),
+}
+
+export const panoFovParam: UrlParam<number | null> = {
+  name: 'fov',
+  defaultValue: null,
+  initialValue: 30,
+  decode: (value) => parseInt(value, 10) ?? null,
```
janjaap, line 119: Same here; return value of `decode` can be `NaN`.

src/app/pages/MapPage/query-params.ts
```diff
@@ -104,11 +104,27 @@ export const baseLayerParam: UrlParam<BaseLayer> = {
...
+  encode: (value) => (typeof value === 'number' ? value.toFixed(0).toString() : null),
+}
+
+export const panoHeadingParam: UrlParam<number | null> = {
+  name: 'heading',
+  defaultValue: null,
+  decode: (value) => parseFloat(value) ?? null,
```
janjaap, line 126: And here

src/app/components/PanoramaViewer/PanoramaViewer.tsx
chrisvanmook: No it wasn't. It is used in a redux state now, but now we adding that in to the URL now: it's easier to maintain and keep the FOV value when sharing the URL

src/app/pages/MapPage/MapPage.tsx
chrisvanmook: Value of the parameters (if decoded from strings) can be of any type. `panoHeading` in this case can be null, meaning it's not set

src/app/pages/MapPage/map-search/PanoramaPreview.tsx
chrisvanmook: I suggest to use the `initialValue` of `panoHeadingParam` then 👍

src/app/pages/MapPage/query-params.ts
chrisvanmook: No, you're right this is not intended. Going to fix this

janjaap (approved):

PR #3974, opened 2021-02-12
[DI2-294] Missing uuid fix

janjaap:
As it turns out, not every node's data, coming the CMS' REST API, has a value for its prop `uuid`. The code was depending on that and thus threw an error for specific nodes. The cause is that related articles don't get the `uuid` prop from the normalization functions. The solution was to remove the `uuid` prop altogether and have all references point the `id` prop instead.

### DIFF
--- a/src/normalizations/cms/normalizeCMSResults.js
+++ b/src/normalizations/cms/normalizeCMSResults.js
@@ -65,14 +65,19 @@ export const getLocaleFormattedDate = ({
   }
 }
 
-export const getLinkProps = ({ type, uuid, field_link, field_special_type, title }, slug) => {
+export const getLinkProps = ({ type, uuid, id, field_link, field_special_type, title }, slug) => {
   let to = {}
 
+  // not every node has a uuid prop
+  const nodeId = uuid || id
+
   if (EDITORIAL_DETAIL_ACTIONS[type]) {
+    const nodeAnchorPropsFn = EDITORIAL_DETAIL_ACTIONS[type]
+
     if (type === CmsType.Special) {
-      to = EDITORIAL_DETAIL_ACTIONS[type](uuid, field_special_type, slug)
+      to = nodeAnchorPropsFn(nodeId, field_special_type, slug)
     } else {
-      to = EDITORIAL_DETAIL_ACTIONS[type](uuid, slug)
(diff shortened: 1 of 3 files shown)


chrisvanmook:
As we continuously having type errors, can we create typings for CMS results and convert this file to typescript?

src/normalizations/cms/normalizeCMSResults.js
```diff
@@ -65,14 +65,19 @@ export const getLocaleFormattedDate = ({
   }
 }
 
-export const getLinkProps = ({ type, uuid, field_link, field_special_type, title }, slug) => {
+export const getLinkProps = ({ type, uuid, id, field_link, field_special_type, title }, slug) => {
   let to = {}
 
+  // not every node has a uuid prop
+  const nodeId = uuid || id
```
jonkoops, line 72: Dear god why Drupal? WHYYYY?

jonkoops, line 72:
```suggestion
  const nodeId = uuid ?? id
```

src/normalizations/cms/normalizeCMSResults.js
```diff
@@ -137,7 +143,7 @@ export const normalizeObject = (data) => {
 
   return {
     key: uuid,
-    id: uuid,
+    id: id || uuid,
```
jonkoops, line 146:
```suggestion
    id: id ?? uuid,
```

jonkoops (approved):

src/normalizations/cms/normalizeCMSResults.js
janjaap: Actually, it's not Drupal. It's our own doing. I'll make some changes, because `uuid` is a prop that isn't needed.

jonkoops: It should be, we use it for articles and such. Doesn't every node in Drupal require a UUID?

jonkoops (approved):

src/normalizations/cms/normalizeCMSResults.js
janjaap: I haven't seen any node with a `uuid` prop in the JSON output.

PR #3988, opened 2021-02-15
Chore increase coverage

janjaap:
This PR:
- fixes an inaccurate check for the `is-object` utility function (and adds tests)
- removes the `isDefined` utility function and its references
- adds tests for the `getGeometry` function
- converts the touched files to TS


src/shared/services/geometry/geometry.ts
```diff
@@ -1,47 +1,60 @@
+import { Point } from 'geojson'
 import isObject from '../is-object'
 import BOUNDING_BOX from '../../../map/services/bounding-box.constant'
 import * as crsConverter from '../coordinate-reference-system/crs-converter'
 
-export const isVestigingAmsterdam = (data) => {
-  let isVestiging = false
+interface GeometryContainer {
```
jonkoops, line 6: This type seems to overlap with the type `ApiAddress` defined in `src/map/types/details.ts`. Perhaps we can use that one instead?

src/shared/services/geometry/geometry.ts
```diff
@@ -1,47 +1,60 @@
...
+  geometrie?: Point
+}
+
+interface Data {
+  bezoekadres?: GeometryContainer
```
jonkoops, line 11: This field is also present in the `PotentialApiResult` type defined in `src/map/types/details.ts`. Perhaps we can use that one instead?

src/shared/services/geometry/geometry.ts
```diff
@@ -1,47 +1,60 @@
...
+  monumentcoordinaten?: Point
+}
+
+export const isVestigingAmsterdam = (geometrie: Point) => {
```
jonkoops, line 15: `isVestigingAmsterdam` seems to be used only in this file, looks like it does not have to be exported.

src/app/components/DataSelection/DataSelectionDownloadButton.tsx
```diff
@@ -33,7 +32,7 @@ const DataSelectionDownloadButton: FunctionComponent<Props> = ({ dataset, active
     }
   })
 
-  if (isDefined(geometryFilter && geometryFilter.markers)) {
+  if (geometryFilter?.markers !== undefined) {
```
jonkoops, line 35: Nice cleanup, gotta love optional chaining :partying_face:

src/shared/services/geometry/geometry.test.ts
```diff
@@ -0,0 +1,57 @@
+import { Point } from 'geojson'
+import * as crsConverter from '../coordinate-reference-system/crs-converter'
+import getGeometry from './geometry'
+
```
jonkoops, line 4: Thanks for writing these tests, it makes it so much clearer what is happening here. :+1:

src/shared/services/geometry/geometry.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    const monumentcoordinaten: Point = { type: 'Point', coordinates: [0.5000001, 0.0000001] }
+
+    const data = {
+      monumentcoordinaten,
+    }
+
+    expect(getGeometry(data)).toEqual(monumentcoordinaten)
+  })
+
+  it('returns geometry from monumentcoordinaten when bezoekadres is out of bounds', () => {
```
jonkoops, line 41: Does this cover all possible types of 'out of bounds' that could occur?

jonkoops (changes requested):

src/shared/services/is-object.test.ts
```diff
@@ -1,13 +1,14 @@
 import isObject from './is-object'
 
 describe('is object', () => {
-  it('should be true for object', () => {
+  it('returns true', () => {
     expect(isObject({ foo: 'bar' })).toBe(true)
```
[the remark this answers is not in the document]
janjaap: `expect(isObject(Object.create(null))).toBe(true)` Will actually be `null` so the assertion is `false`. But I see you point and I'll add a couple more cases.

src/shared/services/is-object.test.ts
```diff
@@ -1,13 +1,14 @@
...
   })
-  it('should be false for null', () => {
+
+  it('returns false', () => {
     expect(isObject(null)).toBe(false)
-  })
-  it('should be false for integer', () => {
     expect(isObject(5)).toBe(false)
+    expect(isObject('')).toBe(false)
+    expect(isObject([])).toBe(false)
```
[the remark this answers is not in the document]
janjaap: Because that is not the intention of the `isObject` function. `null` Is also an object, but not the expected object.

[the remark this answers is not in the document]
jonkoops: Perhaps `isPlainObject` or something like that would be better then? To detect a 'POJO' basically.

src/shared/services/geometry/geometry.ts
janjaap: Whoah, good catch. I've replaced it.

src/shared/services/geometry/geometry.test.ts
janjaap: It doesn't really matter, because it's the return value that matters. And I suspect that the piece of code this test covers, isn't used anywhere.

src/shared/services/is-object.ts
```diff
@@ -0,0 +1,3 @@
+const isObject = (value: any) => value?.constructor?.name === 'Object'
```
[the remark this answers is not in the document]
janjaap: The compiler complained about `value is object` and suggested this:
```typescript
const isObject = (value: any): value is Record<string, unknown> =>
  value?.constructor?.name === 'Object'
```

src/shared/services/is-object.test.ts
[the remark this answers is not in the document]
janjaap: Actually, there is no way to deduct from the code that the intended return value should be false when evaluating an array as input, so I'll add that.

src/app/utils/isObject.test.ts
```diff
@@ -0,0 +1,16 @@
...
+describe('isObject', () => {
+  it('returns true', () => {
+    expect(isObject({ foo: 'bar' })).toBe(true)
+    expect(isObject({})).toBe(true)
+    expect(isObject([])).toBe(true)
+  })
+
+  it('returns false', () => {
+    expect(isObject(null)).toBe(false)
+    expect(isObject(Object.create(null))).toBe(false)
```
jonkoops, line 12: This should be `true` instead, an object without a prototype is still an object.

```suggestion
    expect(isObject(Object.create(null))).toBe(true)
```

src/app/utils/isObject.ts
```diff
@@ -0,0 +1,4 @@
+const isObject = (value: any): value is Record<string, unknown> =>
+  Array.isArray(value) || value?.constructor?.name === 'Object'
```
jonkoops, line 2: This type assertion is incorrect, since we are also checking if the value is an array it could also return an `Array` type. A more correct type would be:

```suggestion
const isObject = (value: any): value is Record<string, unknown> | Array<unknown> =>
  Array.isArray(value) || value?.constructor?.name === 'Object'
```

I would still highly recommended changing this back to `object` and remove the array logic entirely. The reason is that with this type the compiler cannot discriminate the primitive, which means the returned value becomes a union. See for example the `getGeometry` method:

image

Changing the implementation to purely asserting the `object` primitive allows the type to be discriminated properly:

```suggestion
const isObject = (value: any): value is object => value !== null && typeof value === 'object'
```

image

src/shared/services/is-object.ts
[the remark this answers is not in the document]
jonkoops: Looks like a linting error to me, which is a good rule, but in this case it makes no sense.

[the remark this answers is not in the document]
jonkoops: Looks like the linter complaints about this, not the compiler. It's a good rule, but for this type discrimination it's not really applicable.

src/app/utils/isObject.test.ts
janjaap: Changed the condition back to the original and adjusted the tests accordingly

src/app/utils/isObject.ts
janjaap: Reverted to the original implementation

src/shared/services/is-object.ts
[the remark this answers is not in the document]
jonkoops: Updated this so types can be correctly inferred.

jonkoops (approved):

PR #4005, opened 2021-02-23
Using new BAG API for woonplaatsen detail page

chrisvanmook:


src/api/bag/v1/woonplaatsen/detail-config.ts
```diff
@@ -0,0 +1,46 @@
+import { Single as Woonplaatsen } from './types'
```
jonkoops, line 1: I would not put this file in the API directory, this is a change that specifically for the detail panels so let's keep the code there. The API directory is more for direct mapping of API methods and types, without additional transformation.

src/app/pages/MapPage/detail/DetailLinkList.tsx
```diff
@@ -35,7 +35,7 @@ const DetailLinkList: FunctionComponent<DetailLinkListProps> = ({ item, ...other
         if ('url' in link) {
           return (
             <StyledLink
-              key={link.url}
+              key={link.id ?? link.url}
```
jonkoops, line 38: Why would the URL not be unique?

src/app/pages/MapPage/detail/getListFromApi.ts
```diff
@@ -45,12 +45,14 @@ const getListFromApi = (
   }
 
   const results = normalize
-    ? await normalize(response.results || [response])
+    ? // @ts-ignore
+      // eslint-disable-next-line no-underscore-dangle
+      await normalize(response._embedded || response.results || [response])
```
jonkoops, line 50: Can we add some tests for this?

src/map/sagas/nearest-details/nearest-details.js
```diff
@@ -12,7 +12,7 @@ export function* fetchNearestDetails(action) {
     const { uri } = yield call(fetchNearestDetail, location, layers, zoom)
     const isEmbed = yield select(isEmbedded)
 
-    if (uri) {
+    if (uri.type) {
```
jonkoops, line 15: Why `.type` here?

src/map/sagas/nearest-details/nearest-details.test.js
```diff
@@ -36,7 +36,7 @@ describe('fetchNearestDetails', () => {
     },
   }
 
-  it('should call fetchNearestDetails and dispatch the correct actions if uri is returned', () => {
+  it.only('should call fetchNearestDetails and dispatch the correct actions if uri is returned', () => {
```
jonkoops, line 39: This change can be removed.

src/map/services/map-detail/map-detail.ts
```diff
@@ -1,5 +1,6 @@
 import { endpointTypes } from '../map-services.config'
 import environment from '../../../environment'
+import { path as woonplaatsenPath } from '../../../api/bag/v1/woonplaatsen'
```
jonkoops, line 3: Neat way of re-using this info :+1:

jonkoops (changes requested):

src/map/sagas/nearest-details/nearest-details.js
chrisvanmook: we're now always returning an object with values that could also be null from `getDetailPageData` in `actions.js`. We could also return `null` but then we'll have to refactor a lot, so for now I left it like this.

src/app/pages/MapPage/detail/DetailLinkList.tsx
chrisvanmook: results might show twice, not sure why but it's not unique...

jonkoops (approved):

PR #4007, opened 2021-02-23
[DI2-58] Add utility to enable feature toggles on a local machine

jonkoops:


src/app/utils/isFeatureEnabled.ts
```diff
@@ -0,0 +1,6 @@
+export default function isFeatureEnabled(featureName: string) {
+  const enabledFeaturesRaw = window.localStorage.getItem('features')
+  const enabledFeatures = enabledFeaturesRaw ? (JSON.parse(enabledFeaturesRaw) as string[]) : []
```
chrisvanmook, line 3: Perhaps wrap in inside a `try {}  catch` statement, as the value could also be a string?

jonkoops: I've added some code and tests to handle invalid values.

chrisvanmook (approved):

PR #4013, opened 2021-02-24
[DI2-338] Drop support for Firefox ESR versions that are no longer maintained

jonkoops:


jonkoops:
We're going to keep this PR around for the future, as IT services will start phasing out Firefox soon.

chrisvanmook (approved):

public/static/browser-support-check.js
```diff
@@ -0,0 +1,61 @@
...
+var returnPath = getParam('returnPath')
+
+// Go back to the site if the browser is supported.
+if (onNoSupportPath && isSupported && returnPath) {
+  location.replace(returnPath)
+}
+
+// Go the the no support page if the browser is not supported.
+if (!onNoSupportPath && !isSupported) {
+  location.replace(noSupportPath + '?returnPath=' + encodeURIComponent(location.pathname + location.search))
```
alimpens, line 13: Most other url params (or all?) are in Dutch, shouldn't this one be in Dutch as well?

jonkoops: This isn't really meant to be for humans so I think it's fine for this case.

PR #4014, opened 2021-02-24
[DI2-340] Show a warning to users on older versions of Firefox

jonkoops:


src/app/AppBody.tsx
```diff
@@ -85,6 +86,7 @@ const AppBody: FunctionComponent<AppBodyProps> = ({
         <>
           <AppContainer id={APP_CONTAINER_ID} className="main-container">
             <NotificationAlert />
+            <BrowserSupportAlert />
```
janjaap, line 89: Why would you show an alert for something that people do not have control over?

chrisvanmook (approved):

src/app/AppBody.tsx
alimpens: I think an alert is a nice middle ground. Dropping support is a bit much, but showing an alert might motivate people to call ICT, which might help in solving the core issue.

src/app/components/BrowserSupportAlert/BrowserSupportAlert.tsx
```diff
@@ -0,0 +1,32 @@
...
+  // Make sure to update this with the correct version according to the release schedule:
+  // https://wiki.mozilla.org/Release_Management/Calendar
+  if (!isFirefox || version.major >= 78) {
+    return null
+  }
+
+  return (
+    <Alert level="warning" dismissible>
+      Uw browser is verouderd en zal binnenkort niet langer ondersteund worden. Medewerker bij de
+      Gemeente? Zorg dan dat u gebruik maakt van ADW Next.
```
alimpens, line 27: Maybe we should rethink this message. People using TGW's can't use ADW Next.

jonkoops: Yeah, waiting for feedback on this one.

src/app/AppBody.tsx
jonkoops: They have control to pressure IT, that's what we want. They won't listen to us, but I bet they will listen to 3000 angry users.

src/app/components/BrowserSupportAlert/BrowserSupportAlert.tsx
jonkoops: Got approval this is good to go.

PR #4016, opened 2021-02-25
[DI2-58] Add ability to authenticate using Keycloak

jonkoops:


src/shared/services/auth/auth.ts
```diff
@@ -1,356 +1,19 @@
-import joinUrl from '../../../app/utils/joinUrl'
```
jonkoops, line 1: Just FYI all this code has been moved to `auth-legacy.ts`, so no need to review this removal.

src/shared/services/auth/auth-legacy.ts
```diff
@@ -0,0 +1,356 @@
+import joinUrl from '../../../app/utils/joinUrl'
```
jonkoops, line 1: Just FYI all this code has been moved from the original auth code, this code hasn't changed so it doesn't have to be reviewed.

src/shared/services/auth/auth-keycloak.ts
```diff
@@ -0,0 +1,64 @@
...
+  url: environment.KEYCLOAK_URL,
+  realm: environment.KEYCLOAK_REALM,
+  clientId: environment.KEYCLOAK_CLIENT,
+})
+
+export function getAccessToken() {
+  return keycloak.token ?? ''
+}
+
+export function login() {
```
chrisvanmook, line 14: Why not directly use `keycloak.login`? Is this because we replace the `legacy-auth` file?

src/shared/services/auth/auth-keycloak.ts
```diff
@@ -0,0 +1,64 @@
...
+export function logout() {
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  keycloak.logout()
+}
+
+export async function initKeycloak() {
+  const authenticated = await keycloak.init({
+    checkLoginIframe: false,
+    pkceMethod: 'S256',
+    onLoad: 'check-sso',
```
chrisvanmook, line 28: Can we link this to the documentation on why we need to use these values?

src/shared/services/auth/auth-keycloak.ts
jonkoops: Yes, for now we only replace things at a low level. This will be replaced with Keycloak methods in the future.

src/shared/services/auth/auth-keycloak.ts
jonkoops: Sure, I've added a link to the documentation.

chrisvanmook: Perhaps also documentation to an amsterdam doc so we know why we have to use the specific `pkceMethod`?

chrisvanmook (approved):

PR #4039, opened 2021-03-12
[DI2-368] Fixed legend login link color

alimpens:


src/map/components/legend/MapLegend.jsx
```diff
@@ -31,6 +31,18 @@ const NonSelectableLegendParagraph = styled.p`
   margin-bottom: ${themeSpacing(2)};
 `
 
+const Notification = styled.div`
```
jonkoops, line 34: Is this something we could possibly replace with the [Alert](https://amsterdam.github.io/amsterdam-styled-components/?path=/docs/ui-alert--variants) component?

jonkoops (approved):

chrisvanmook (approved):

src/map/components/legend/MapLegend.jsx
alimpens: Yes, that makes sense, I'll change it

alimpens: I replaced the styled div with the Alert component. This does mean that the notification is focusable now, without it being interactive (only the link inside the notification is interactive). Is that ok, a11y-wise?

jonkoops (approved):

src/map/components/legend/MapLegend.jsx
jonkoops: Yeah, this makes sense. It can be focused so that the screen reader can read it when navigating around.

PR #4046, opened 2021-03-17
Release v3.26.0

alimpens:


jonkoops:
Couple of notes:
- I see that there is now `v3.25.0` commit on `develop`, perhaps we should first merge `master` back into `develop` as it already exists there?
- I am missing the [git tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging), NPM will create these for you when you run `npm version`, you can push them with `git push --tags`

jonkoops (approved):

PR #4052, opened 2021-03-24
[DI2-366] Added woonplaatsen to dataselection table

chrisvanmook:


jonkoops (approved):
LGTM! It might be a good idea to add both 'Amsterdam' and 'Weesp' as prefilled values, so we don't have to rely on the API response.

PR #4053, opened 2021-03-24
Added woonplaatsen to dataselection table links on openbare ruimte de…

chrisvanmook:
…tail page

### DIFF
--- a/src/map/services/map-services.config.ts
+++ b/src/map/services/map-services.config.ts
@@ -24,6 +24,7 @@ import {
   DetailResultNotification,
   ExtraApiResults,
   InfoBoxProps,
+  InternalLink,
   PotentialApiResult,
 } from '../types/details'
 import adressenNummeraanduiding from './adressen-nummeraanduiding/adressen-nummeraanduiding'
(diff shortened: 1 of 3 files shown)


src/map/services/map-services.config.ts
```diff
@@ -283,65 +283,78 @@ export const getLocationDefinitionListBlock = (result: any, gridArea: string): D
...
+      infoBox: getInfoBox({
+        description: GLOSSARY.DEFINITIONS.VESTIGING.description,
+        url: GLOSSARY.DEFINITIONS.VESTIGING.url,
+        plural: GLOSSARY.DEFINITIONS.VESTIGING.plural,
+      }),
+      links: [
+        {
+          to: {
+            pathname: config[DataSelectionType.HR].path,
+            search: `${PARAMETERS.VIEW}=volledig&${PARAMETERS.FILTERS}=${filterParamValue}`,
```
jonkoops, line 332: Perhaps we can use `URLSearchParams` here to encode the params properly?

src/map/services/map-services.config.ts
```diff
@@ -283,65 +283,78 @@ export const getLocationDefinitionListBlock = (result: any, gridArea: string): D
...
+      infoBox: getInfoBox({
+        description: GLOSSARY.DEFINITIONS.OBJECT.description,
+        url: GLOSSARY.DEFINITIONS.OBJECT.url,
+        plural: GLOSSARY.DEFINITIONS.OBJECT.plural,
+      }),
+      links: [
+        {
+          to: {
+            pathname: config[DataSelectionType.BRK].path,
+            search: `${PARAMETERS.VIEW}=volledig&${PARAMETERS.FILTERS}=${filterParamValue}`,
```
jonkoops, line 350: This seems to be consistently the same as in the other entries, perhaps we can move this into a variable?

src/map/services/map-services.config.ts
```diff
@@ -283,65 +283,78 @@ export const getLocationDefinitionListBlock = (result: any, gridArea: string): D
...
+): DetailResultItemLinkList[] => {
+  const filterParamValue = JSON.stringify(
+    filters.reduce(
+      (acc, { key, value }) => ({
+        ...acc,
+        [key]: value,
+      }),
+      {},
+    ),
+  )
```
jonkoops, line 300: Might be cleaner to use [Object.fromEntries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) here (untested):
```suggestion
  const filterParamValue = JSON.stringify(
    Object.fromEntries(filters.map(({ key, value }) => [key, value])),
  )
```

jonkoops (approved):
Looks good to me, but check out my comments as they might be useful.

PR #4055, opened 2021-03-26
Fixed bug with showing all map layers and map collections (DI2-209)

alimpens:


chrisvanmook (approved):

jonkoops (approved):

src/store/middleware/preserveUrlParametersMiddleware.test.ts
```diff
@@ -26,7 +26,19 @@ describe('Custom Redux Middleware', () => {
     expect(getParametersForRouteMock).not.toHaveBeenCalled()
   })
 
-  it('should use the custom middleware when pathname does not includes "kaart"', () => {
+  it('should use the custom middleware when pathname includes "kaarten"', () => {
+    // @ts-ignore
+    window.location = {
```
chrisvanmook, line 31: can't you do:
Object.defineProperty(window, 'location', {
    value: {
      ...window.location,
      pathname: '/kaarten/foo/bar',
      search: '?foo=bar',
    },
  })

chrisvanmook (approved):

chrisvanmook (approved):

src/app/pages/MapPage/detail/buildDetailUrl.test.tsx
```diff
@@ -0,0 +1,35 @@
+import buildDetailUrl from './buildDetailUrl'
+
+describe('return detail panel path', () => {
+  it('should return a path starting with "/kaart/" when pathname includes "kaart"', () => {
+    const locationSpy = jest.spyOn(window, 'location', 'get').mockReturnValue({
+      ...window.location,
```
jonkoops, line 6: You might get away without adding `window.location` here

src/app/pages/MapPage/detail/buildDetailUrl.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+      type: 'foo',
+      subtype: 'bar',
+      id: 'baz',
+    }
+
+    const location = buildDetailUrl(params)
+
+    expect(location.pathname).toBe('/data/foo/bar/baz/')
+  })
+})
```
jonkoops, line 35: I'm missing a test case for the `kaarten` route.

jonkoops (approved):

src/app/pages/MapPage/detail/buildDetailUrl.test.tsx
alimpens: You get a type error if you omit `window.location`, the `Location` type expects all `window.location` props. I could overwrite the type I guess, but the current approach might be cleaner?

src/app/pages/MapPage/detail/buildDetailUrl.test.tsx
alimpens: Good catch, I'll add it

src/app/pages/MapPage/detail/buildDetailUrl.test.tsx
jonkoops: If you want to pass only a subset of the fields in your type you can cast it using the `as` keyword. In this case that would be something like:

```ts
 const locationSpy = jest.spyOn(window, 'location', 'get').mockReturnValue({ pathname: '/kaart', search: '?foo=bar' } as Location)
```

alimpens: Nice, changed it

jonkoops (approved):

PR #4092, opened 2021-04-14
Converted CMS normalisations to TS

chrisvanmook:


src/app/components/HomePage/AboutBlock.tsx
```diff
@@ -96,51 +91,49 @@ const AboutBlock = () => {
...
-                ))
-              : null}
+            {isFulfilled(resultAboutData) &&
+              resultAboutData.value.map((aboutData, index) => (
+                <StyledCardColumn
+                  wrap
+                  key={aboutData.key || index}
+                  span={{ small: 1, medium: 2, big: 3, large: 3, xLarge: 3 }}
+                >
+                  <AboutCard loading={isPending(resultAboutData)} {...aboutData} />
```
jonkoops, line 109: I really hate this idea of spreading props from a model over a component. I'd much rather see it being passed directly as a single prop. For example `value={aboutData}`.

src/app/components/HomePage/AboutBlock.tsx
```diff
@@ -96,51 +91,49 @@ const AboutBlock = () => {
...
-                ))
-              : null}
+            {isFulfilled(resultAbout) &&
+              resultAbout.value.map((about, index) => (
+                <StyledCardColumn
+                  wrap
+                  key={about.key || index}
+                  span={{ small: 1, medium: 2, big: 3, large: 3, xLarge: 3 }}
+                >
+                  <AboutCard loading={isPending(resultAbout)} {...about} />
```
jonkoops, line 134: Same here, this spreading seems like a dumb thing.

jonkoops, line 134: The loading state is not going to work here as we are already in a `isFulfilled` state.

src/app/components/HomePage/AboutBlock.tsx
jonkoops: I left a TODO comment to resolve this at a later point in time.

src/app/components/HomePage/AboutBlock.tsx
jonkoops: I left a TODO comment to resolve this at a later point in time.

src/app/components/HomePage/AboutBlock.tsx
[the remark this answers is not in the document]
jonkoops: Pushed a fix for this so it matches the original logic, but there is still room for improvement. I've added a TODO comment to resolve this later.

src/app/components/HomePage/AboutBlock.tsx
jonkoops: Pushed a fix for this so it matches the original logic, but there is still room for improvement. I've added a TODO comment to resolve this later.

src/app/components/HomePage/AboutCard.tsx
```diff
@@ -42,33 +43,27 @@ const StyledLink = styled(Link)`
...
+  teaser,
+  intro,
+  linkProps,
+}) => (
   <StyledLink {...linkProps} variant="blank">
     <StyledCard backgroundColor="level2" shadow isLoading={loading}>
       <StyledCardContent>
         <Heading as="h3">{shortTitle || title}</Heading>
-        <Paragraph dangerouslySetInnerHTML={{ __html: teaser || intro }} />
+        {teaser || (intro && <Paragraph dangerouslySetInnerHTML={{ __html: teaser || intro }} />)}
```
jonkoops, line 63: This will actually result in a bug, as `teaser` will always be `undefined` when passed into the paragraph.

jonkoops (changes requested):

src/app/components/HomePage/AboutCard.tsx
jonkoops: Resolved this one.

jonkoops (approved):

PR #4124, opened 2021-04-30
Added feature toggle beta map

chrisvanmook:


src/store/middleware/isIgnoredPath.ts
```diff
@@ -1,6 +1,10 @@
 // These paths are ignored by all the shitty Redux code.
 // Add your route here if you are moving away from Redux First Router.
-const ignoredPaths = ['kaart', 'bouwdossiers']
+import { FEATURE_BETA_MAP, isFeatureEnabled } from '../../app/features'
+
+const ignoredPaths = ['bouwdossiers']
+
+ignoredPaths.push(isFeatureEnabled(FEATURE_BETA_MAP) ? 'data' : 'kaart')
```
jonkoops, line 7: Not sure if this correct, we should only ignore the `kaart` path right? Since the `data` path still uses all of the old reducers and such, and not React Router.

jonkoops (changes requested):

jonkoops (approved):

PR #4157, opened 2021-05-04
Add related articles to publications

alimpens:


alimpens:
Doesn't work yet, because the related articles seem to be removed from the cms results [here](https://github.com/Amsterdam/atlas/blob/d605d8eafa55cc8720902044b7b9ab02b8c80a50/src/shared/services/cms/cms-json-api-normalizer.ts#L55)

PR #4199, opened 2021-05-14
Replace parameter constants with UrlParams

jonkoops:


src/map/services/map-services.config.ts
```diff
@@ -313,7 +314,7 @@ export const getShowInTableBlock = (
   )
   const search = new URLSearchParams({
     [PARAMETERS.VIEW]: 'volledig',
```
chrisvanmook, line 316: Could also be replaced with the new `viewParam`

jonkoops: Good catch. I'll open up a PR to convert the remaining parameters.

PR #4235, opened 2021-05-20
Max zoom on auto-centre on map, permanent tooltip on polyline + enabl…

chrisvanmook:
…e / disable features via URL params


src/app/index.tsx
```diff
@@ -11,6 +11,17 @@ import { disableFeature, enableFeature, getEnabledFeatures } from './features'
 import resolveRedirects from './redirects'
 import './sentry'
 
+// Able to enable and disable feature-toggles via URL parameters
```
jonkoops, line 14: Let's not do this, I'd rather build an actual UI. Right now the console method is excessively documented as well.

jonkoops (changes requested):

jonkoops (approved):

PR #4306, opened 2021-06-03
Clean up README and remove redundant sections

jonkoops:


README.md
```diff
@@ -1,24 +1,20 @@
-# Data en informatie
```
chrisvanmook, line 1: We should call this `Data en Informatie / Dataportaal` as `Data en Informatie` is the name of the application and used outside Datapunt. `Dataportaal` is still the more technical term....

chrisvanmook (approved):

PR #4352, opened 2021-06-10
Download complete dossier 

thomasm0:
This feature introduces the following changes to the Construction Dossier Pages:
- A download all button ('Download geheel dossier')
- A checkbox to toggle all documents in Document Details
- If the download all button is used but the user does not have rights to view all Documents, a new modal (RestrictedFilesModal) appears before the Download modal listing the excluded files.

For testing, this dossier is completely public: 
- data/bouwdossiers/bouwdossier/ST9024/

These dossiers have some public + some restricted: 
- data/bouwdossiers/bouwdossier/XY5800/
- data/bouwdossiers/bouwdossier/XY10950/

This dossier is completely restricted:
- data/bouwdossiers/bouwdossier/SA88871/

---
- 7537 Downloaden volledig dossier


src/app/pages/ConstructionDossierPage/components/DocumentDetails/DocumentDetails.test.tsx
```diff
@@ -220,15 +261,72 @@ describe('DocumentDetails', () => {
...
+        onRequestLoginLink={() => {}}
+        onDownloadFiles={() => {}}
+      />,
+      { wrapper: wrapperWithToken },
+    )
+
+    expect(screen.queryByTestId('allFilesToggle')).not.toBeInTheDocument()
+  })
+
+  it('checking the select all checkbox updates the label accordingly', () => {
```
thomasm0, line 310: With this test - initially I would have like to tested if all the Document checkboxes are checked, however, in this test the FilesGallery component uses a mock component. If you include the real component there's quite a few console warnings from FilesGallery - "Warning: An update to IIIFThumbnail inside a test was not wrapped in act". It's not breaking but certainly not very clean.

I've seen similar warnings with the download modal (it seems to attempt API requests in tests) but didn't identify an easy fix. I'll continue exploring but it doesn't seem a major issue.

src/app/pages/ConstructionDossierPage/components/DossierDetails/DossierDetails.test.tsx
```diff
@@ -1,42 +1,68 @@
...
 import DossierDetails from './DossierDetails'
 
 jest.mock('../FilesGallery')
-jest.mock('../LoginLinkRequestModal')
 
 const FilesGalleryMock = mocked(FilesGallery)
-const LoginLinkRequestModalMock = mocked(LoginLinkRequestModal)
 
-describe.skip('DossierDetails', () => {
+const VALID_DECODED_TOKEN: DecodedToken = {
```
thomasm0, line 15: Would be nice to move this valid token handling to a shared util as it's also used in AuthTokenContext

jonkoops: Yeah for sure, although making tests DRY is not really that big of an issue for me.

src/app/pages/ConstructionDossierPage/components/RestrictedFilesModal/RestrictedFilesModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+const ModalBody = styled.div`
+  padding: ${themeSpacing(6, 4)};
+`
+
+const RestrictedFilesModal: FunctionComponent<RestrictedFilesModalProps> = ({
+  files,
+  onClose,
+  onDownload,
+}) => {
+  const fileList = files.map(({ filename, url }) => <ListItem key={url}>{filename}</ListItem>)
```
jonkoops, line 32: Since this variable is only used once, why not inline this logic in the `<List>`?

src/app/pages/ConstructionDossierPage/components/RestrictedFilesModal/RestrictedFilesModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+          />
+        </Heading>
+      </TopBar>
+      <Divider />
+      <ModalBody>
+        <>
+          <Paragraph>
+            U heeft niet de bevoegdheden om de volgende beveiligde documenten te downloaden:
+          </Paragraph>
+          <List variant="bullet">{fileList}</List>
```
jonkoops, line 55: This list can be quite long depending on the dossier, perhaps this is something we should check the worst case scenario of to see if the UX still holds up.

src/app/pages/ConstructionDossierPage/components/DossierDetails/DossierDetails.tsx
```diff
@@ -28,7 +38,11 @@ const DossierDetails: FunctionComponent<DossierDetailsProps> = ({
   dossier,
   ...otherProps
 }) => {
-  const [showLoginLinkRequestModal, setShowLoginLinkRequestModal] = useState(false)
+  const [selectedFiles, setSelectedFiles] = useState<Bestand[]>([])
+  const [currentModal, setCurrentModal] = useState<string | null>(null)
```
jonkoops, line 42: It could be a good idea to keep a type for all possible modals types. For example:

```ts
export type ModalType = 'login' | 'select' | 'download' | 'restricted'

const [currentModal, setCurrentModal] = useState<ModalType | null>(null)
```

This will prevent typos from sneaking in which might cause bugs. For example, `setCurrentModal('ownload') ` would produce an error when compiling since it's not a valid value.

src/app/pages/ConstructionDossierPage/components/DossierDetails/DossierDetails.tsx
```diff
@@ -38,17 +52,90 @@ const DossierDetails: FunctionComponent<DossierDetailsProps> = ({
...
+  // Only allow downloads from a signed in user if authenticated with Keycloak.
+  // TODO: This logic can be removed once we switch to Keycloak entirely.
+  const disableDownload = isAuthenticated() && !isFeatureEnabled(FEATURE_KEYCLOAK_AUTH)
+
+  // Check user rights by the primary dossier object's access prop or if every child document is 'RESTRICTED'
+  const restricted =
+    dossier.access === 'RESTRICTED' ||
+    dossier.documenten.every((doc) => doc.access === 'RESTRICTED')
+
+  const hasRights = useMemo(() => {
```
jonkoops, line 64: Since `restricted` is only used to determine `hasRights` perhaps the logic for it can be moved into this `useMemo`.

jonkoops, line 64: On a sidenote, this logic is now duplicated between `DossierDetails` and `DocumentDetails`, perhaps we can put it in a seperate hook somehow and re-use it?

src/app/pages/ConstructionDossierPage/components/DossierDetails/DossierDetails.tsx
```diff
@@ -38,17 +52,90 @@ const DossierDetails: FunctionComponent<DossierDetailsProps> = ({
...
+    let files: Array<Bestand> = []
+
+    // If user doesn't have full permissions check if we have to exclude any files
+    if (!scopes.includes(SCOPES['BD/X'])) {
+      files = dossier.documenten
+        .filter((doc) => doc.access === 'RESTRICTED')
+        .map((doc) => doc.bestanden)
+        .flat()
+
+      setRestrictedDownloadFiles(files)
```
jonkoops, line 93: This is the only place this value is set, which makes me suspect this could also be a computed value using `useMemo`.

```ts
const restrictedFiles = useMemo(() => {
  // Logic to determine restricted files here.
}, [])
```

src/app/pages/ConstructionDossierPage/components/DossierDetails/DossierDetails.tsx
```diff
@@ -38,17 +52,90 @@ const DossierDetails: FunctionComponent<DossierDetailsProps> = ({
...
+    }
+
+    return files
+  }
+
+  function handleDownloadAllClick() {
+    // For each dossier document - get the files and then merge each document's results into a single array
+    const files = dossier.documenten
+      .filter((doc) => {
```
jonkoops, line 102: Perhaps it would be a good idea to extract this logic into a function that can be re-used throughout the component. For example:

```ts
function hasAccess(scopes: string[], access: BouwdossierAccess) {
  if (access === 'PUBLIC') {
    return true
  }

  return scopes.includes(SCOPES['BD/X']) 
}
```

This should make it easier to use later on, for example:

```ts
const files = dossier.documenten
  .filter((doc) => hasAccess(scopes, doc.access))
  .map((doc) => doc.bestanden)
  .flat()
```

jonkoops (changes requested):

src/app/pages/ConstructionDossierPage/components/RestrictedFilesModal/RestrictedFilesModal.tsx
thomasm0: Agreed, I did a quick check to see what happens if there are more list items than the modal's max-height (75%) and it becomes scrollable - it's not the most intuitive in terms of UX but I think it's sufficient.

src/app/pages/ConstructionDossierPage/components/DocumentDetails/DocumentDetails.tsx
```diff
@@ -42,17 +42,33 @@ const GalleryContainer = styled.div`
...
+  position: relative;
+`
+
+const StyledCheckbox = styled(Checkbox)`
+  padding: ${themeSpacing(3)};
+
+  // Todo: fix in ASC
+  input {
+    opacity: 0;
+    left: 12px;
```
chrisvanmook, line 55: you could use `themeSpacing(3)` here (3 * 4 = 12px), for consistency sakes :)

src/app/pages/ConstructionDossierPage/components/DocumentDetails/DocumentDetails.tsx
```diff
@@ -42,17 +42,33 @@ const GalleryContainer = styled.div`
...
   dossierId,
   dossier,
   document,
   onRequestLoginLink,
+  onDownloadFiles,
 }) => {
   const [selectedFiles, setSelectedFiles] = useState<Bestand[]>([])
-  const [showSelectFilesModal, setShowSelectFilesModal] = useState(false)
-  const [showRequestDownloadModal, setShowRequestDownloadModal] = useState(false)
+  const [allFilesSelected, setAllFilesSelected] = useState(false)
```
chrisvanmook, line 69: I believe we could introduce a derived state with `useMemo` from `selectedFiles` from this, instead of having to set it's own state.

src/app/pages/ConstructionDossierPage/components/DossierDetails/DossierDetails.tsx
thomasm0: This is awesome! I briefly looked at string enums thinking this was possible somehow but I guess I got distracted/took the easy String option. Thanks!

chrisvanmook (approved):

PR #4375, opened 2021-06-14
Feature/rating component

chrisvanmook:


package-lock.json
```diff
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "data-verkenner",
```
jonkoops, line 8: NPM adds this if the directory name is different from the name of the package. To avoid this rename your workspace to `data-verkenner` as well. Otherwise this will keep showing up in the diff.

jonkoops (approved):
I love that we can start splitting up the map config mess! LGTM!

PR #4382, opened 2021-06-16
Convert MapLegend to typescript

thomasm0:
Here's the initial commit with a working version of the MapLegend in Typescript. I've had to add a few `// @ts-ignore` comments in the MapLegend and also make one change to the LegendItem type, which I will run through in comments.

Primary issue is that there seems to be some conflicts between the mapLayer and legendItem vars/types - hopefully I've just used the wrong type.

Also still need to write a test for this

### DIFF
--- a/src/app/pages/MapPage/legacy/panel/MapLegend.tsx
+++ b/src/app/pages/MapPage/legacy/panel/MapLegend.tsx
@@ -3,17 +3,22 @@ import { Alert, Checkbox, Icon, Label, styles, themeColor, themeSpacing } from '
 import { useMatomo } from '@datapunt/matomo-tracker-react'
 import classNames from 'classnames'
 import queryString from 'querystring'
+import type { ParsedUrlQueryInput } from 'querystring'
+import type { FunctionComponent } from 'react'
 import { createRef, useEffect, useMemo, useState } from 'react'
-import styled, { css } from 'styled-components'
+import styled from 'styled-components'
 import LoginLink from '../../../../components/Links/LoginLink/LoginLink'
 import { useIsEmbedded } from '../../../../contexts/ui'
 import SearchPlus from '../../../../../shared/assets/icons/search-plus.svg'
 import MAP_CONFIG from '../services/map.config'
 import { isAuthorised } from '../utils/map-layer'
+import type { MapLayer } from '../services'
+import type { MapPanelOverlay } from './MapPanel'
 
 const TitleWrapper = styled.div`
   display: flex;
 `
+
 const StyledLabel = styled(Label)`
   width: 100%;
 `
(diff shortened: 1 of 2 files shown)


src/app/pages/MapPage/legacy/panel/MapLegend.tsx
```diff
@@ -243,6 +266,7 @@ const MapLegend = ({
             {!isEmbedView ? (
               <StyledCheckbox
                 className="checkbox"
+                // @ts-ignore
```
thomasm0, line 269: I get a notice that the name property isn't recognized ("No overload matches this call.")

src/app/pages/MapPage/legacy/panel/MapLegend.tsx
```diff
@@ -295,6 +319,7 @@ const MapLegend = ({
                         <StyledCheckbox
                           id={mapLayer.id}
                           className="checkbox"
+                          // @ts-ignore
                           variant="tertiary"
```
thomasm0, line 323: I get a notice that the variant property isn't recognized ("No overload matches this call."). Looks like the property may have been removed in a recent commit but I'm not sure if we used this same version: Amsterdam/amsterdam-styled-components@43ca777

src/app/pages/MapPage/legacy/panel/MapLegend.tsx
```diff
@@ -56,34 +61,47 @@ const LayerButton = styled.div.attrs({
...
       isOpen &&
-      css`
+      `
         transform: rotate(180deg);
       `}
   }
 `
 
-const constructLegendIconUrl = (mapLayer, legendItem) => {
+const constructLegendIconUrl = (mapLayer: MapLayer, legendItem: MapLayer) => {
```
thomasm0, line 70: When I inspected the mapLayer and legendItem objects in browser they seemed to be the same. Howver, if I use MapLayerLegendItem for legendItem I get errors:

Property 'layers' does not exist on type 'MapLayerLegendItem'.
  Property 'layers' does not exist on type 'LegendItem'.

thomasm0 (commented):
I also notice on line 170 MapPanel.tsx

`const [isOpen, setOpen] = useState(isEmbedView ?? false)`

Shouldn't this be [isOpen, setIsOpen]? I guess the above is just a typo?

src/app/pages/MapPage/legacy/panel/MapLegend.tsx
jonkoops: Yeah this was removed and can be ommitted

src/app/pages/MapPage/legacy/panel/MapLegend.tsx
jonkoops: Yeah, it can be omitted

jonkoops (approved):

PR #4452, opened 2021-06-30
Feature/osd multiple images

thomasm0:
This feature enables the Dossier ImageViewer to handle multiple images. It adds the following features:

- Prev/next button to browser through documents
- Enables the [OpenSeaDragon reference strip](https://openseadragon.github.io/examples/ui-reference-strip/)


src/app/pages/ConstructionDossierPage/components/ImageViewer/ImageViewer.tsx
```diff
@@ -50,34 +70,72 @@ const ImageViewer: FunctionComponent<ImageViewerProps> = ({
   const [loading, setLoading] = useState(true)
   const [error, setError] = useState(false)
   const [viewer, setViewer] = useState<Viewer>()
-  const fileExtension = fileName.split('.').pop()
+  const [selectedFileIndex, setSelectedFileIndex] = useState<number>(
```
jonkoops, line 73: Could also be a `useMemo` based on `selectedFileName`

jonkoops (approved):

src/app/pages/ConstructionDossierPage/components/ImageViewer/ImageViewer.tsx
thomasm0: I think that was my original setup, however, the OSD reference strip doesn't provide any event on image change - so I hook into the parent OSD viewer object and use this to update the index if the 'page' event happens (on page changes), so that the ImageViewer filename + download file can update accordingly.

PR #4459, opened 2021-07-06
Update header search value on search dropdown menu click

thomasm0:
I've added a Context/Provider to the HeaderSearch setup to accordingly update the search input value and URL term query parameter on AutoSuggestItem link clicks. I also cleaned up the Header file structure:

- `src/header` files moved into `src/app/components/Header` directory
- `src/store/redux-first-router` had 2 files `actions.ts` (two util functions) and `actions.test.ts` - As we no longer use redux, I've separated the 2 functions and moved them into utils (`getDetailPageData.ts` and `getIdEndpoint.ts`). I renamed `extractIdEndpoints` to `getIdEndpoints` to align with other util filenames and added a test.


src/app/components/Header/HeaderSearchContext.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <HeaderSearch />
+        </HeaderSearchProvider>,
+      ),
+    )
+  })
+
+  it('provides an empty value if the URL param is empty', () => {
+    useParamMock.mockReturnValue(['', () => {}])
+
+    const Tests = () => {
```
chrisvanmook, line 37: You should rather write a test like this with the `renderHook` util. Then we don't need to render a test component.

```suggestion
    const { result } = renderHook(() => useHeaderSearch(), {
      wrapper: ({ children }) => <HeaderSearchProvider>{children}</HeaderSearchProvider>,
    })
    
    expect(result.current.searchInputValue).toEqual('')
```

src/app/components/Header/HeaderSearchContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+  null,
+)
+
+export default HeaderSearchContext
+
+const HeaderSearchProvider: FunctionComponent = ({ children }) => {
+  const [searchQuery] = useParam(queryParam)
+  const [searchInputValue, setSearchInputValue] = useState<string>(searchQuery)
+
+  const updateSearchInputValue = (query: string) => {
```
chrisvanmook, line 24: Why did you wrap this in a separate method and not rename `setSearchInputValue` to `updateSearchInputValue`?

src/app/utils/getDetailPageData.ts
```diff
@@ -1,18 +1,10 @@
...
-}
-
 const URL_SUBTYPES_MAPPING = {
   openbareruimtes: 'openbareruimte',
   woonplaatsen: 'woonplaats',
   ligplaatsen: 'ligplaats',
 }
 
-export const getDetailPageData = (endpoint: string) => {
+export default (endpoint: string) => {
```
chrisvanmook, line 7: I think we should always name default exports, otherwise they will show up as anonymous in stack traces

src/app/components/Header/HeaderSearchContext.tsx
thomasm0: That is a very good question that I'm not sure the answer to

src/app/components/Header/HeaderSearchContext.test.tsx
thomasm0: Aha, this is very good to know. I was a bit confused with the renderHook

chrisvanmook (approved):

PR #4543, opened 2021-07-28
Optimize Panorama view layout

thomasm0:


src/app/pages/MapPage/MapPage.tsx
```diff
@@ -171,6 +194,12 @@ const MapPage: FunctionComponent = () => {
     mapInstanceRef.current?.setZoom(zoom)
   }, [zoom])
 
+  useMemo(() => {
```
thomasm0, line 197: It seems like this should be a useEffect but whenever I try with useEffect you start getting issues with the mini-map positioning being wrong. I looked through the useEffect order to see if the map is getting repositioned by something else but cannot see anything.

chrisvanmook: let's look into this because this should be in a useeffect indeed

thomasm0: For reference - `flyTo` can be used to panTo a location and zoom (first argument latlng, second arg is zoom level)

For example `mapInstanceRef.current.flyTo(position, 12)`

thomasm0: The following changes to the MapPage.tsx useEffects seems to work, however, @chrisvanmook maybe you know some more specific map links where the zoom level is predefined. I haven't looked at the param handling with this either:

```
useEffect(() => {
  if (mapInstanceRef.current) {
    mapInstanceRef.current.invalidateSize()
  }
}, [panoActive, panoFullScreen, panoPitch, mapInstanceRef])

useEffect(() => {
  if (mapInstanceRef.current && position) {
    if (panoFullScreen) {
      mapInstanceRef.current.flyTo(position, 12)
    } else {
      mapInstanceRef.current.panTo(position)
    }
  }
}, [panoActive, panoFullScreen])

useEffect(() => {
  // mapInstanceRef.current?.setZoom(zoom)
}, [zoom])

useEffect(() => {
  if (mapInstanceRef && panoFullScreen) {
    setZoom(12)
  }
}, [panoFullScreen])
```

chrisvanmook (approved):

PR #4574, opened 2021-08-11
Display menu control and infobar in panorama full screen

thomasm0:
I've added the 'Meest recent' button and ViewerInfoBar to the panorama full screen. I've created a new component `PanoramaFullScreenMenuControl` to ease the styling changes. The `ViewerInfoBar` component required quite a few style changes to prevent the mini-map conflicting with its positioning - particuarly if the data panel is active.

I'm open to design changes on this one too, this seemed like the easiest way to fit things in.

Screenshot 2021-08-11 at 17 16 54
Screenshot 2021-08-11 at 17 17 01
Screenshot 2021-08-11 at 17 17 08
Screenshot 2021-08-11 at 17 17 13


src/app/components/ViewerInfoBar/ViewerInfoBar.tsx
```diff
@@ -23,14 +23,65 @@ const ViewerInfoBarStyle = styled.div`
...
+          border-right: 1px solid rgba(255, 255, 255, 0.2);
+        }
+      }
+    `}
+
+  ${({ displayPortrait, panelActive }) =>
+    displayPortrait &&
+    panelActive &&
+    css`
+      span:last-of-type {
```
lpsmitamsterdam, line 60: i would probably try to avoid a display: none if possible. maybe you could make the spans into two separate styled components and render them based on the props? this way you avoid having to pass the props into the container divs. You could also compose the the styles that way.  Of course this is a matter of preference but it might reduce the amount of styles you have to write in this case.

thomasm0:
Moved the context menu to the top right and some other responsive styling improvements

Screenshot 2021-08-16 at 17 16 09

chrisvanmook (approved):

PR #4575, opened 2021-08-12
Feature/scss autosuggest refactor

lpsmitamsterdam:


src/app/components/Header/auto-suggest/AutoSuggest.tsx
```diff
@@ -22,6 +22,41 @@ const NoResults = styled(Paragraph)`
...
+  }
+
+  @media screen and (min-width: 769px) {
+    max-width: calc(100% - 45px);
+  }
+`
+
+const AutoSuggestTip = styled.h3`
+  font-size: 14px; // was font-size($xs-font) (scaled)
+  color: ${themeColor('tint', 'level5')}; //was $secondary-gray60 !note not the same but closest;
```
lpsmitamsterdam, line 55: @chrisvanmook @thomasm0 how shall we approach the mismatch between colors in the sass theme and the asc? should we maybe refine this together with design?

chrisvanmook: as discussed, let's leave it for now, we can create another ticket and PR's to fix this and allign with design

src/app/components/Header/auto-suggest/AutoSuggestItem.tsx
```diff
@@ -43,9 +43,29 @@ export interface AutoSuggestItemProps {
...
+  padding-bottom: calc(${10 / 2}px); // was $base-whitespace
+  padding-top: calc(${10 / 3}px); // was $base-whitespace
+  position: relative;
+  text-align: left;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  width: 100%;
+  text-decoration: none;
+  &:hover,
+  &--active {
```
chrisvanmook, line 65: is `--active` being used anywhere?

chrisvanmook (approved):

src/app/components/Header/auto-suggest/AutoSuggestItem.tsx
lpsmitamsterdam: good you mention this actually, as useTraverselist & headersearch refer to the active class so i have to fix that as the classname has changed

src/app/components/Header/auto-suggest/AutoSuggest.tsx
```diff
@@ -22,6 +22,41 @@ const NoResults = styled(Paragraph)`
...
+  position: absolute;
+  right: 45px; + 10px * 0.5;
+  max-height: calc(100vh - 160px); // 160 is max height of the header
+  overflow: auto;
+
+  // Makes sure the autosuggest box starts behind the SearchBar
+  top: 45px - 5px;
+
+  @media screen and ${breakpoint('min-width', 'tabletM')} {
+    top: 45px; // This is the height of the SearchBarMenu from @amsterdam/asc-ui
```
thomasm0, line 42: I noticed 5px more top spacing in this version than compared to production. On production this media-query only applies the styling `max-width: calc(100% - 45px)`

thomasm0 (changes requested):
Everything looks good except for the AutoSuggestDropdown 45px instead of 40px

thomasm0 (approved):

PR #4587, opened 2021-08-17
Feature/scss table refactor

lpsmitamsterdam:
replaces the table from the scss with styled component from asc,


lpsmitamsterdam:
also renamed the page as Chris suggested

src/app/pages/UpdatesPage/UpdatesPage.tsx
```diff
@@ -0,0 +1,71 @@
...
+`
+
+const UpdatesPage: FunctionComponent = () => (
+  <ContentContainer>
+    <Container>
+      <Row>
+        <Column span={{ small: 1, medium: 2, big: 6, large: 12, xLarge: 12 }}>
+          <div>
+            <Heading>Actualiteit</Heading>
+            <div className="qa-page">{renderContents()}</div>
```
chrisvanmook, line 33: I know it probably was like this already, but I think we can put the content of `renderContents` directly into this div. Also not sure if we need `qa-page`, if it's not used anywhere let's remove it

thomasm0 (approved):

PR #4593, opened 2021-08-18
Update @amsterdam/use-promise to version 3

jonkoops:


chrisvanmook:
@jonkoops Did you got this to work in other projects that use jest?

jonkoops:
@chrisvanmook I wanted to spend some time figuring out what is causing the issue that complains about ES modules. By all accounts this should work, bit haven't found the time to work on it.

chrisvanmook:
Yeah I've tried to add this to the `transformIgnorePatterns` but had no effect.. Will have a better look into this later

jonkoops:
That was my guess as well, however I think it will also have to be included in the tsconfig in order to be transformed (if the TypeScript compiler indeed does compile this). I also tried to enable ES modules in Jest but this is still too experimental.

PR #4601, opened 2021-08-19
Cleanup sass files and fixed minor style issues

chrisvanmook:


index.ejs
```diff
@@ -40,8 +40,15 @@
     <% } %>
 
     <style>
-      .root {
+      html, body, .root, .root > div {
         height: 100%;
+      }
+
+      body {
```
lpsmitamsterdam, line 47: should this be in here or in global style?

lpsmitamsterdam (approved):

index.ejs
chrisvanmook: I think it's better to have this here, then we don't have a flash from white background to grey when the javascript / style is loaded

PR #4648, opened 2021-09-20
Feature/custom map layers options

thomasm0:


lpsmitamsterdam (approved):

src/environment.ts
```diff
@@ -6,6 +6,7 @@ const environment = {
   CMS_ROOT: process.env.CMS_ROOT || 'https://acc.cms.data.amsterdam.nl/',
   GRAPHQL_ENDPOINT:
     process.env.GRAPHQL_ENDPOINT || 'https://acc.api.data.amsterdam.nl/cms_search/graphql/',
+  MAP_ROOT: process.env.MAP_ROOT || 'https://acc.map.data.amsterdam.nl/',
```
chrisvanmook, line 9: IMHO we can hardcode this, as this would be part of a (temporary) feature toggle / POC. otherwise we'd have to also add this in the openstack repo

PR #4881, opened 2021-11-18
Removed all CMS content from data verkenner

chrisvanmook:


.env.example
```diff
@@ -1,7 +1,6 @@
 DEPLOY_ENV=acceptance
 IIIF_ROOT=https://acc.images.data.amsterdam.nl/
 API_ROOT=https://acc.api.data.amsterdam.nl/
-CMS_ROOT=https://acc.cms.data.amsterdam.nl/
```
thomasm0, line 4: bye bye 👋

PR #4909, opened 2021-11-23
Feature/cypressimproved

chrisvanmook:


cypress/regressionSuite/staticPage-colofonfooterlink.ts
```diff
@@ -0,0 +1,148 @@
+/* eslint-disable cypress/no-unnecessary-waiting */
```
chrisvanmook, line 1: I don't think it's a good idea to ignore this rule. Instead we should wait on requests, instead of using `cy.wait`:  https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Intercepts

cypress/regressionSuite/staticPage-colofonfooterlink.ts
```diff
@@ -0,0 +1,148 @@
...
+describe('Static Page for Footer module', () => {
+  beforeEach(() => {
+    cy.viewport(1536, 960)
+  })
+
+  describe('Static page footer module, without authenicated', () => {
+    describe(`Colofon Navigation Block `, () => {
+      beforeEach(() => {
+        cy.visit('/')
+        cy.wait(3000)
```
chrisvanmook, line 21: This can also be removed when using intercepts

cypress/regressionSuite/staticPage-overoisheaderlink.ts
```diff
@@ -0,0 +1,179 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+/* eslint-disable cypress/no-unnecessary-waiting */
```
chrisvanmook, line 2: Again, we should intercept the requests in this test instead of using `.wait`'s

cypress/regressionSuite/staticPage-overoisheaderlink.ts
```diff
@@ -0,0 +1,179 @@
...
+  describe('Over OIS module in HomePage, without authenicated', () => {
+    describe(`Header Over OIS Navigation Block`, () => {
+      beforeEach(() => {
+        cy.visit('/')
+        cy.wait(3000)
+
+        cy.get(ABOUTOISDROPDOWN.overoismenuCheck).last().click()
+      })
+
+      it('should have link to the about ris', () => {
```
chrisvanmook, line 27: "ris" typo?

cypress/smokeSuite/homePage-colofonfooterlink.ts
```diff
@@ -0,0 +1,118 @@
+/* eslint-disable cypress/no-unnecessary-waiting */
```
chrisvanmook, line 1: Again, we should intercept the requests in this test instead of using .wait's

PR #4968, opened 2021-12-09
Enable exhaustive-deps

chrisvanmook:


lpsmitamsterdam (approved):

thomasm0:
@lpsmitamsterdam There was an issue with the PanoramaViewer where the controls wouldn't work. It looks like other useEffect statements overrule it - I'm thinking something to do with history/location changes. 

Anyway this should be fixed in my last commit, it's not the ideal solution (just an ignore for the warning) but I see that there's other `eslint-disable-next-line react-hooks/exhaustive-deps` lines in other Map related files so 🤞 this is enough

PR #5215, opened 2022-02-14
New mobile navigation and search menus

thomasm0:


lpsmitamsterdam:
Schermafbeelding 2022-02-28 om 13 04 28

src/shared/components/Header/HeaderMenu.tsx
```diff
@@ -1,200 +1,16 @@
-import { ChevronRight } from '@amsterdam/asc-assets'
-import { Link, MenuButton, MenuFlyOut, MenuInline, MenuItem, MenuToggle } from '@amsterdam/asc-ui'
```
lpsmitamsterdam, line 2: did we remove this from the secure asc already? I think we wont need it anymore when its removed from here right?

src/App.tsx
```diff
@@ -189,16 +188,8 @@ const App: FunctionComponent = () => {
...
               <AppBar data-testid="header">
                 <AppLogo />
-                <HeaderSearchProvider>
-                  <HeaderSearch />
-                </HeaderSearchProvider>
-                <HeaderMenu type="default" data-testid="header-menu-default" showAt="laptopM" />
-                <HeaderMenu
-                  align="right"
-                  type="mobile"
-                  data-testid="header-menu-mobile"
```
lpsmitamsterdam, line 199: maybe also remove the old cypress tests and selectors for this menu? or should we discuss with jayashree first

src/shared/components/Header/MobileHeaderMenu.tsx
```diff
@@ -0,0 +1,82 @@
...
+    width: auto;
+  }
+`
+
+const MobileHeaderMenu: FunctionComponent = () => {
+  const [activeMobileMenu, setActiveMobileMenu] = useState<null | string>(null)
+  const width = useWindowWidth()
+
+  // Prevent the main app body scrolling in the background
+  // Be sure to include document.body to prevent annoying scroll jumps on iOS keyboard open/close
```
lpsmitamsterdam, line 26: so we need it on both, on just the body isnt working?

src/shared/components/Header/MobileHeaderNav.tsx
```diff
@@ -0,0 +1,60 @@
...
+import type { FunctionComponent } from 'react'
+import styled, { css } from 'styled-components'
+import MobileHeaderNavLinks from './MobileHeaderNavLinks'
+
+const StyledButton = styled(Button)<{ active: boolean }>`
+  padding: 0 ${themeSpacing(6)};
+  width: 114px;
+  height: 100%;
+  color: #3b3b3b;
+  border-left: 1px solid #d4d4d4;
```
lpsmitamsterdam, line 12: these colors cannot come from asc?

src/shared/components/Header/HeaderMenu.tsx
thomasm0: I see that they're still in the secure asc index.ts. Unfortunately they're still used in the desktop navigation menu `src/shared/components/Header/HeaderNav.tsx`

src/App.tsx
thomasm0: Good point, I don't think there's any specific mobile cypress tests for navigation yet. I see the existing ones fail too but I'm not entirely sure which cypress directory is the latest. Let's merge the `feature/cypressimproved` PR first and then we can ask @jeswaran2021 to write some tests

src/shared/components/Header/MobileHeaderNav.tsx
thomasm0: No they aren't in the colors var

src/shared/components/Header/MobileHeaderMenu.tsx
thomasm0: Yeah it's a weird one (big thanks for Apple for moving the address bar to the bottom 🤬).

So if you don't include both `document.documentElement` and `document.body` on iphone (iphone 12 pro max; iOS 15.2) the following bug occurs:
1. Touch search button to expand search menu
2. Touch search input to activate keyboard
3. Touch 'Menu' to open the navigation menu

You cannot close the menu as the scroll jumps (seems to be triggered when the keyboard automatically closes)

simulator_screenshot_9F018408-151F-4778-AE88-E97194A2B530

lpsmitamsterdam (approved):

PR #5319, opened 2022-03-29
Feature/api bag v1

thomasm0:
This branch contains the following DSO APIs:
- bag/nummeraanduiding
- gebieden/wijk

`src/pages/MapPage/components/MapPanel/MapPanel.tsx` file has a function (`shouldUseDetailPanelV2`) that will route the above two endpoints to use `src/pages/MapPage/components/DetailPanelV2` and other requests use the current production setup.

I've tried to structure things similarly to the old setup but with cleaner code and an easier separation between building components vs API fetch requests + the related normalization. I've considered each API endpoint that we serve directly via a route as as service (for example: https://data.amsterdam.nl/data/bag/nummeraanduiding/0363200012172982/ = the service `bag/nummeraanduiding`), which are all listed in `src/services/index.ts`. 

A service has the following properties:
- **endpoint** - the URL path to the API endpoint
- **endpointParams** - (optional) any query parameters to append to the fetch request URL, for example: `?_expandScope=...`
- **childRequests** - (optional) Any secondary requests we know we need before rendering DetailItems/secondary fetch requests dependent on a property from the initial parent fetch request (for example - verblijfsobject ID)
- **getMapGeometry** - For building the geometry and location object to position the map correctly
- **prepareDetails** - To render each `DetailItem` and any associated logic/fetch requests. I tried to keep any fetch requests minimal/out of this step but with verblijfsobject we don't know some required values without completing the verblijfsobject request, which happens in the `childRequests` step.

I imagine some of these can be simplified/refactored so we don't need to define this on every service - for example, with `getMapGeometry` but I imagine this will become clearer as we implement other endpoints.

### Overall flow
1. User loads /data/bag/nummeraanduiding/0363200012151123
2. `src/pages/MapPage/components/MapPanel/MapPanel.tsx` sees `/bag/nummeraanduiding` so knows to render `src/pages/MapPage/components/DetailPanelV2`.
3. `src/pages/MapPage/components/DetailPanelV2` renders and calls `getDetailData`
4. `getDetailData` checks if a service definition exists for the requested endpoint (it returns null if nothing is returned)
5. `getDetailData` calls `getData`, which makes a fetch request to the parent endpoint and also each request present in `childRequests`
6. `getDetailData` calls `getMapGeometry`
7. `getDetailData` calls `getDetails`, which gets the DetailItems ready to render
8. React renders the components

### To improve
Naming and type definitions also definitely need improving; `Services` as the directory sounds a big vague, likewise with `apis` (which we already have a directory for)

### Current issues
- `FetchAndSplitListData` and `FetchAndLegacySplitListData` make duplicate requests (as the component rerenders, as well as multiple parents)


lpsmitamsterdam:
Schermafbeelding 2022-04-01 om 10 21 28
Schermafbeelding 2022-04-01 om 10 20 54

im wondering why we actually have a link to nummeraanduidingen on the nummeraanduidingen detail page, on prod it just jumps and does nothing, in your branch it shows this 'geen detailweergave beschikbaar'

lpsmitamsterdam:
Schermafbeelding 2022-04-01 om 10 21 28
Schermafbeelding 2022-04-01 om 10 20 54

src/pages/MapPage/components/DetailPanelV2/DetailPanelV2.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const results = usePromise(() => getDetailData(), [type, subType, id])
+
+  useAsyncMapPanelHeader(
+    results,
+    isFulfilled(results) ? results.value?.data?.subTitle : null,
+    isFulfilled(results) ? results.value?.data?.title : null,
+  )
+
+  if (isRejected(results)) {
```
lpsmitamsterdam, line 86: maybe its out of scope for this PR but we could choose to keep this out of the detail panel and keep it as a separate  component?

lpsmitamsterdam: without the ternary for the content

lpsmitamsterdam: without the ternary for the content

lpsmitamsterdam:
Schermafbeelding 2022-04-01 om 10 53 54

i see various requests are made twice, also occurs on prod though, is this something you looked into? can we avoid this?

src/pages/MapPage/components/MapPanel/MapPanel.tsx
```diff
@@ -38,6 +39,10 @@ const LegendDrawerPanelHeader = styled(DrawerPanelHeader)`
   min-height: 50px;
 `
 
+const shouldUseDetailPanelV2 = (path: string) =>
```
lpsmitamsterdam, line 42: thinking about the naming, maybe we should clarify this is about implementing the new api?

lpsmitamsterdam, line 42: thinking about the naming, maybe we should clarify this is about implementing the new api?

thomasm0: I'll change to this something along the lines of `shouldUseDSOApi`

PR #5320, opened 2022-03-30
Feature/idle time logout

thomasm0:
The test for this is weak, testing this is quite difficult, as `react-idle-timer` works with service workers so won't work natively with Jest/react-testing-library. v5 is on the way with a release candidate available now that supposedly makes testing easier but I couldn't get it to work.


src/shared/components/IdleTimeLogout/IdleTimeLogout.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const IdleTimeLogout: FunctionComponent<IdleTimeLogoutProps> = ({ timeout }) => {
+  const [modalOpen, setModalOpen] = useState(false)
+
+  const handleIdle = () => {
+    setModalOpen(true)
+    document.body.style.overflow = 'hidden'
+
+    setTimeout(() => {
+      logout()
```
lpsmitamsterdam, line 39: im thinking, not sure what are the requirements exactly but shouldnt we give the user a button in that modal to stay active?

lpsmitamsterdam (approved):

PR #5682, opened 2022-10-07
Carousel for dataplatform service page

kiruys:
Carousel:
- occurs on mobile screen width and shows the dataplatform services
- snaps to nearest slide when scrolling
- indicates slide index by the 'colored' links
- scrolls to the slide when clicking the associated link


thomasm0 (approved):
Screenshot 2022-10-18 at 09 48 44

Looks good! I noticed one thing with the carousel - if I select the last element (by clicking on the `CarouselLinks`) the active prop only applies to the first and second items, not the last.

PR #5741, opened 2022-11-07
Feature/fontsizes in rem

kiruys:


rabiexxx:
@kiruys  While the above implementation is perfectly fine and has nothing wrong with it. Yet it's very difficult to implement ( which you already did ) and maintain as it is not very easily readable and will require either a calculator or a cheat sheet on the side. 
A nice trick to avoid the hassle of working rem calculations is to set the root element font-size to the equivalent of 10px in rem units, now since rem units are relative to the root element, each 1 rem unit equals 10px. So let's say a paragraph element which should be 14px could be then written as 1.4rem. Which is much more readable and easier to work with. 
```

html { font-size: 62.5%; } /* =10px */
body { font-size: 1.4rem; } /* =14px */
h1 { font-size: 2.4rem; } /* =24px */
```

If you are curious to know more, here is an [article](https://www.sitepoint.com/understanding-and-using-rem-units-in-css/) which talks about this

kiruys:
> @kiruys While the above implementation is perfectly fine and has nothing wrong with it. Yet it's very difficult to implement ( which you already did ) and maintain as it is not very easily readable and will require either a calculator or a cheat sheet on the side. A nice trick to avoid the hassle of working rem calculations is to set the root element font-size to the equivalent of 10px in rem units, now since rem units are relative to the root element, each 1 rem unit equals 10px. So let's say a paragraph element which should be 14px could be then written as 1.4rem. Which is much more readable and easier to work with.
> 
> ```
> 
> html { font-size: 62.5%; } /* =10px */
> body { font-size: 1.4rem; } /* =14px */
> h1 { font-size: 2.4rem; } /* =24px */
> ```
> 
> If you are curious to know more, here is an [article](https://www.sitepoint.com/understanding-and-using-rem-units-in-css/) which talks about this

Thanks Rabih, for looking at my PR and the comment. That seems like a good option! What remains to be calculated is the relative line-height. Also, the new design system uses sizes with easier calculations and often the same numbers. So it's not that difficult.

thomasm0 (approved):

PR #5763, opened 2022-11-11
Remove product page and references to it

kiruys:
see https://dev.azure.com/CloudCompetenceCenter/Data%20Verkenner/_workitems/edit/60974


kiruys:
I don't get why the `over OIS` cypress test is suddenly failing. But a next ticket includes the removal of the over ois page, so I decided to remove it here to solve the issue.

PR #5777, opened 2022-11-17
Add user role translations

kiruys:


src/shared/components/Header/UserScopeModal.tsx
```diff
@@ -28,18 +31,49 @@ const StyledHeading = styled(Heading)<{ as?: string }>`
...
+  flex-direction: row;
+  align-items: center;
+  p {
+    margin: 0;
   }
 `
 
+const StyledRole = styled(StyledParagraph)`
+  margin: 0 0 0 2.5rem;
+  color: #767676;
```
lpsmitamsterdam, line 74: should the colors not come from some variable?

kiruys: done

thomasm0 (approved):

